PR #1, opened 2015-04-01
Multithreading

DwayneJengSage:
As far as you're all concerned, this is the initial version. This is the refactored, parallelizable (using multiple threads in a single JVM process and using multiple JVMs / EC2 machines for each upload date), modular, and hopefully readable by someone other than me.

This compiles, but still needs to go through a round of testing.


src/main/java/org/sagebionetworks/bridge/exporter/BridgeExporter.java
```diff
@@ -0,0 +1,276 @@
...
+            if (RECORD_LIMIT > 0 && numTotal > RECORD_LIMIT) {
+                break;
+            }
+
+            String recordId = oneRecordKey.getString("id");
+            try {
+                // re-query health data records to get values
+                Item oneRecord;
+                try {
+                    oneRecord = recordTable.getItem("id", recordId);
```
marcel-blonk, line 169: This here will slow this down quite a bit and should be unnecessary. The query itself can return the values already

src/main/java/org/sagebionetworks/bridge/synapse/SynapseHelper.java
```diff
@@ -0,0 +1,157 @@
...
+
+public class SynapseHelper {
+    public static final Map<String, ColumnType> BRIDGE_TYPE_TO_SYNAPSE_TYPE =
+            ImmutableMap.<String, ColumnType>builder()
+                    .put("attachment_blob", ColumnType.FILEHANDLEID)
+                    .put("attachment_csv", ColumnType.FILEHANDLEID)
+                    .put("attachment_json_blob", ColumnType.FILEHANDLEID)
+                    .put("attachment_json_table", ColumnType.FILEHANDLEID)
+                    .put("boolean", ColumnType.BOOLEAN)
+                    .put("calendar_date", ColumnType.STRING)
```
marcel-blonk: There is a date type

src/main/java/org/sagebionetworks/bridge/worker/ExportWorker.java
```diff
@@ -0,0 +1,69 @@
...
+ * Generic export worker. This class encapsulates basic configuration, helpers, and clients needed by export workers
+ * and serves as the parent class for both Synapse export workers and iOS survey export workers.
+ */
+public abstract class ExportWorker extends Thread {
+    // Configured externally
+    private ExportWorkerManager manager;
+    private String studyId;
+
+    // Internal state
+    private final LinkedBlockingQueue<ExportTask> taskQueue = new LinkedBlockingQueue<>();
```
marcel-blonk, line 18: Probably want to limit the size of the queue so you don't run out of memory if takers are slow

src/main/java/org/sagebionetworks/bridge/exporter/BridgeExporter.java
DwayneJengSage, line 169: To clarify, we need to re-query DDB because the uploadDate-index only projects the keys, not the whole row. That said, we could probably benefit from some batching here.

Querying DDB isn't the bottleneck (either in throughput, latency, or capacity), but this is something we definitely want to address if we want to scale up to 100 studies. I've filed a JIRA https://sagebionetworks.jira.com/browse/BRIDGE-458

src/main/java/org/sagebionetworks/bridge/worker/ExportWorker.java
DwayneJengSage, line 18: Right now, the health data record table is only 800mb for 3 weeks worth of data, so I don't think memory will be a concern. For scaling 100x, we should consider moving the task queue to SQS. I've also included this in JIRA https://sagebionetworks.jira.com/browse/BRIDGE-458

PR #5, opened 2016-01-28
update dependencies, configs, minor bugfix, add Travis config

DwayneJengSage:
- update dependencies
- add SQS queues and record ID override S3 buckets to config
- fix a big getting Study Info from DDB
- add Travis config

Testing done:
- mvn verify (unit tests, coverage, FindBugs)
- tested locally
- set up Dev env in Elastic Beanstalk and tested it there

### DIFF
--- a/src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
+++ b/src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
@@ -175,7 +175,8 @@ public class DynamoHelperTest {
         DynamoHelper helper = new DynamoHelper();
         helper.setDdbStudyTable(mockStudyTable);
 
-        // execute and validate
+        // execute and validate - studyInfo is null because the StudyInfo builder returns null if either attributes are
+        // null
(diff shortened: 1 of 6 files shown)


src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
```diff
@@ -161,4 +162,38 @@ public void getStudyInfo() {
...
+        Table mockStudyTable = mock(Table.class);
+        when(mockStudyTable.getItem("identifier", "test-study")).thenReturn(studyItem);
+
+        // set up Dynamo Helper
+        DynamoHelper helper = new DynamoHelper();
+        helper.setDdbStudyTable(mockStudyTable);
+
+        // execute and validate
+        StudyInfo studyInfo = helper.getStudyInfo("test-study");
+        assertNull(studyInfo);
```
alxdarksage, line 181: If I remember, this is null because the builder returns null if the study doesn't have these two configuration keys? Maybe a note here

alxdarksage:
Looks good, can merge when you're ready

PR #6, opened 2016-02-16
use columns defined in Synapse; add data groups

DwayneJengSage:
Two changes here:
1. Currently, if the Synapse table contains fields different from what Bridge-EX expects, Bridge-EX will error out. This makes it difficult to add new fields to Bridge-EX. This change updates Bridge-EX to use the fields define in the Synapse table and ignore any fields the Synapse table doesn't understand. This means we can add new fields to Bridge-EX and still be backwards compatible with older tables and older studies.
2. Adds data groups to Bridge-EX.

Testing done:
- mvn verify passes (unit tests, Jacoco coverage, FindBugs)
- manual tests: (1) old table w/o dataGroups, (2) manually add dataGroups field, (3) create a new table w/ dataGroups field


src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -190,12 +276,79 @@ private String initSynapseTable(ExportTask task) throws BridgeExporterException,
...
+        rowValueMap.put("healthCode", record.getString("healthCode"));
+        rowValueMap.put("externalId", BridgeExporterUtil.sanitizeDdbValue(record, "userExternalId", 128, recordId));
+
+        // Data groups, if present. Sort them in alphabetical order, so they appear consistently in Synapse.
+        Set<String> dataGroupSet = record.getStringSet("userDataGroups");
+        if (dataGroupSet != null) {
+            List<String> dataGroupList = new ArrayList<>();
+            dataGroupList.addAll(dataGroupSet);
+            Collections.sort(dataGroupList);
+            rowValueMap.put("dataGroups", DATA_GROUP_JOINER.join(dataGroupList));
```
alxdarksage, line 307: DATA_GROUP_JOINER.join(Sets.newTreeSet(dataGroupSet)) would be 1-2 lines

DwayneJengSage, line 307: Discussed this with Alx over IM. I feel like compact code is _less_ readable, not more, so I'm not particularly inclined to make this change.

PR #8, opened 2016-03-08
Status Table; refactor table creation

DwayneJengSage:
Two changes:
1. At the end of export, Bridge-EX will now write to a "status table" in each Synapse project (creating the table if it doesn't exist). The sole purpose of this table is to signal any automated jobs to start (notably, Chris Bare's data analysis that's being consumed for mPower viz). As such, the only thing each row in this table contains is the date of the last Bridge-EX execution for that study. (Example, Bridge-EX runs on 2016-03-07, processes data for 2016-03-06, writes 2016-03-07 to the status table.)
2. Moved table creation code to SynapseHelper. It's no longer a Handler exclusive thing, so we're moving it further down the stack.
# File-By-File Overview
- AppVersionExportHandler - SynapseMetaTables is now shared between AppVersion and Status tables, so we're moving these constants to SynapseHelper.
- SynapseExportHandler - Refactoring a lot of the logic to shared classes. The logic for writing to the DDB table map is now in ExportWorkerManager, since it depends heavily on stuff that's already there. Creating columns, tables, and ACLs is moved to SynapseHelper. (The diff is a bit messy. get/setSynapseTableIdFromDdb is moved to ExportWorkerManager, createTable() is the replacement for getColumnNameListForNewTable().)

This has an interesting side effect that creating tables now returns the Synapse table ID instead of the list of column IDs, so we'll need to make another call Synapse to get the column IDs back. This isn't a big deal, since it only happens when a new table is created, which is very rare. (Part of this includes renaming getColumnNameListForExistingTable() to simply getColumnNameListForTable(), since it will be called for both.)
- SynapseHelper - Added createTableWithColumnsAndAcls(). This encapsulates common operations around table creation, notably creating column models and setting ACLs. There's no retry annotation on this method because the underlying methods have retries, and retries on multiple layers generally lead to weird things happening.
- SynapseStatusTableHelper - Helper to write to Synapse Status table, or create it if necessary. The main method calls two private helper methods:
1. createStatusTable(), which creates the table (calling through to SynapseHelper), then writes the table ID to DDB (calling through to ExportWorkerManager).
2. writeStatus(), which calls Synapse to get the column IDs, then writes a single row with a single column to the Status table. There's some validation here in case the Status table is malformed. It's not future-proof, but we can make it future-proof if/when we decide to expand the usage of the Status table.
- ExportTask - Each task now needs to know which study IDs it operated on, so we can write to the appropriate Status tables. ExportTask already keeps track of task state (TSVs, outstanding tasks), so adding a set of Study IDs to it should be a no-brainer.
- ExportWorkerManager - 3 major changes here:
1. Moving get/setSynapseTableIdFromDdb from SynapseExportHandler to here, so that it can be used by SynapseStatusTableHelper. (Note that there is a circular dependency between ExportWorkerManager and SynapseStatusTableHelper.) 
2. In addSubtaskForRecord() (which is called for every Health Data Record), we add the study ID to the task. As previously noted, this is to keep track of which Status tables to update.
3. In endOfStream() (which is called at the end of every export task), after we're done uploading to all health data tables and all app version tables, we get the set of study IDs and call SynapseStatusTableHelper to write to the respective status tables. We do this in a try-catch so one failure to write to a status table doesn't fail the whole job.

Plus 2 minor changes:
1. Remove "final" from some of these methods. These are getters with no side effect, so there shouldn't be a risk of removing final. Plus, removing final allows us to mock/spy these methods if needed.
2. Remove getDdbClient(). Since get/setSynapseTableFromDdb was removed from SynapseExportHandler, this getter is no longer needed. (If in the future, it is needed, it's trivial to add it back.)
# Testing Done
- Added and updated unit tests.
- mvn verify (unit tests, jacoco coverage checks, findbugs)
- manual tests: (1) Tested in a blank Synapse project to make sure all tables were created properly. (2) Tested again in the same project to make sure we can write to existing Synapse projects.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
@@ -1,0 +1,141 @@
+package org.sagebionetworks.bridge.exporter.synapse;
+
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.sagebionetworks.client.exceptions.SynapseException;
+import org.sagebionetworks.repo.model.table.ColumnModel;
+import org.sagebionetworks.repo.model.table.ColumnType;
+import org.sagebionetworks.repo.model.table.PartialRow;
+import org.sagebionetworks.repo.model.table.PartialRowSet;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import org.sagebionetworks.bridge.exporter.exceptions.BridgeExporterException;
+import org.sagebionetworks.bridge.exporter.worker.ExportTask;
+import org.sagebionetworks.bridge.exporter.worker.ExportWorkerManager;
+
+/**
+ * Highly specialized helper, which writes the status row to the Synapse status table, creating it if it doesn't
+ * already exist.
+ */
+@Component
+public class SynapseStatusTableHelper {
+    // Constants package-scoped to be available to unit tests.
+    static final String COLUMN_NAME_UPLOAD_DATE = "uploadDate";
+    static final List<ColumnModel> COLUMN_LIST;
+    static {
+        // Construct column definition list. This table is only used as a flag to signal that we're done uploading, so
+        // we only need one column with the date.
+        // NOTE: ColumnType.DATE is actually a timestamp. There is no calendar date type.
+        ColumnModel uploadDateColumn = new ColumnModel();
+        uploadDateColumn.setName(COLUMN_NAME_UPLOAD_DATE);
+        uploadDateColumn.setColumnType(ColumnType.STRING);
+        uploadDateColumn.setMaximumSize(10L);
+
+        COLUMN_LIST = ImmutableList.of(uploadDateColumn);
+    }
+
+    private ExportWorkerManager manager;
+    private SynapseHelper synapseHelper;
+
+    /**
+     * Export worker manager, used for its many utility methods, like get/setSynapseTableIdFromDdb, data access team,
+     * principal ID, parent project ID, etc.
+     */
+    @Autowired
+    final void setManager(ExportWorkerManager manager) {
+        this.manager = manager;
+    }
+
+    /** Synapse Helper, used to create the Synapse table. */
+    @Autowired
+    final void setSynapseHelper(SynapseHelper synapseHelper) {
+        this.synapseHelper = synapseHelper;
+    }
+
+    /**
+     * Writes the export task status to the status table for the given export task and study. If the status table
+     * doesn't exist, this will create it.
+     *
+     * @param task
+     *         export task to write status for
+     * @param studyId
+     *         study ID to write status for
+     * @throws BridgeExporterException
+     *         if there's an unusual condition, like mismatched table columns
+     * @throws InterruptedException
+     *         if a Synapse asynchronous request thread is interrupted
+     * @throws SynapseException
+     *         if the Synapse call fails
+     */
+    public void initTableAndWriteStatus(ExportTask task, String studyId) throws BridgeExporterException,
+            InterruptedException, SynapseException {
+        // Does the table already exist? If not, create it.
+        String synapseTableId = manager.getSynapseTableIdFromDdb(task, SynapseHelper.DDB_TABLE_SYNAPSE_META_TABLES,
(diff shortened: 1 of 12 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
```diff
@@ -0,0 +1,141 @@
...
+    static final String COLUMN_NAME_UPLOAD_DATE = "uploadDate";
+    static final List<ColumnModel> COLUMN_LIST;
+    static {
+        // Construct column definition list. This table is only used as a flag to signal that we're done uploading, so
+        // we only need one column with the date.
+        // NOTE: ColumnType.DATE is actually a timestamp. There is no calendar date type.
+        ColumnModel uploadDateColumn = new ColumnModel();
+        uploadDateColumn.setName(COLUMN_NAME_UPLOAD_DATE);
+        uploadDateColumn.setColumnType(ColumnType.STRING);
+        uploadDateColumn.setMaximumSize(10L);
```
alxdarksage, line 35: This will never include a time component, e.g. updating more than once per calendar day? I'm assuming this is 10 for YYYY-MM-DD

alxdarksage, line 35: How/where is Chris going to record what data he has processed? Is it easier for end users like researchers if it's in the same status table? (Not something you need to act on, just wondering.)

src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
```diff
@@ -0,0 +1,141 @@
...
+                    ", expected 1 column, got " + columnList.size() + " columns");
+        }
+
+        // We know there is exactly one column in the status table.
+        ColumnModel column = columnList.get(0);
+        String colName = column.getName();
+        if (!COLUMN_NAME_UPLOAD_DATE.equals(colName)) {
+            throw new BridgeExporterException("Wrong column in status table for study " + studyId + ", expected " +
+                    COLUMN_NAME_UPLOAD_DATE + ", got " + colName);
+        }
```
alxdarksage, line 126: Can you be more lenient and just look for the column with the correct name + type? Then if someone wants to piggyback further information about data processing on the table, presumably they could do that. Creating new rows you could follow existing schema and only write to the one column you care about.

alxdarksage:
Can merge whenever you're ready.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
DwayneJengSage, line 126: We intentionally separate Bridge-EX tables from researcher tables. This is to prevent researchers from modifying the original data (accidentally or intentionally). This follows the same model as all the other tables Bridge-EX has created.

There's also not really any point for researchers to add information to this table. This table is intended entirely to signal automated processes that Bridge-EX has finished running for the day. If researchers need to add their own data about a day's worth of data, there's nothing stopping them from creating their own tables.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
DwayneJengSage, line 35: We don't ever include a time component, because that wouldn't be useful. (That and the Synapse timestamp column isn't an ISO timestamp and is difficult to query on, so we want to avoid that anyway.) The idea is to signal that Bridge-EX has finished running for a given day. Whether that's at 2:15am or 2:17am is inconsequential.

We do have plans to make Bridge-EX run more frequently (perhaps hourly), but this is such an in-depth change across multiple systems that what we do here doesn't really have an impact.

As for how/where Chris is recording his results, that's outside the scope of Bridge-EX. Currently, Bridge-EX only exports data from Bridge to Synapse. We may have need in the future to process Synapse data back into Synapse, but that's another project that's far outside the scope of this change.

As for whether the data should be written back into the Status Table: In the specific case, no, because the data Chris is generating is not keyed off of upload date. In the general case, see my other response.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseStatusTableHelper.java
alxdarksage, line 126: Not intending researchers or data, just metadata from other processes that are involved in something like the visualization pipeline. Like Chris' script for example.

PR #11, opened 2016-04-20
automatically create new columns when the schema changes

DwayneJengSage:
This is needed to support mutable schemas. Bridge-EX will now compare the existing columns in a Synapse table against the column defs generated from a schema, and update the Synapse table accordingly if they are different. Has checks to prevent against column deletion and column modification. Allows column re-ordering. Optimized to not update tables if no columns are added.

This check and update is only run once per Exporter run.

Validation done:
- added unit tests
- Jacoco coverage checks
- findbugs checks

Manual tests:
- created a Synapse table with extra fields and wrong field types and sizes; verified the "deleted column" and "modified column" checks in the logs
- created a Synapse table with missing columns (including both health data columns and common columns) and columns out of order; verified that BridgeEX creates and reorders columns appropriately
- created a Synapse table with only swapped columns; verified that BridgeEX doesn't update the table if there are no new columns

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
@@ -115,6 +115,7 @@ public abstract class SynapseExportHandler extends ExportHandler {
         try {
             // get TSV info (init if necessary)
             TsvInfo tsvInfo = initTsvForTask(task);
+            tsvInfo.checkInitAndThrow();
 
             // Construct row value map. Merge row values from common columns and getTsvRowValueMap()
             Map<String, String> rowValueMap = new HashMap<>();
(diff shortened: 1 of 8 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -155,53 +160,129 @@ private synchronized TsvInfo initTsvForTask(ExportTask task) throws BridgeExport
...
-        return synapseTableId;
+        // Get the column list.
+        return synapseHelper.getColumnModelsForTableWithRetry(synapseTableId);
     }
 
-    // Helper method to get the column name list for a Synapse table. This queries the column model list from Synapse.
-    private List<String> getColumnNameListForTable(String synapseTableId) throws SynapseException {
-        // Get columns from Synapse
+    // Helper method to detect when a schema changes and updates the Synapse table accordingly. Will reject schema
+    // changes that delete or modify columns. Optimized so if no columns were inserted, it won't modify the table.
```
alxdarksage, line 210: What happens if it rejects a delete or a type change? The export fails, or is the export later put in a new table? It looks like it just skips those columns on export... ?

PR #13, opened 2016-04-26
prefer column max size in Synapse over column max size in generated column defs

DwayneJengSage:
There is a bug in the semi-mutable Synapse tables code. We check to make sure we don't modify any columns (the only fields we care about being type and max size).

Unfortunately, really old versions of BridgeEX created tables with size 1000 (and in other cases, we manually recreated columns with much smaller sizes), so the check isn't backwards compatible. Instead, if the sizes are different, we use the size from the Synapse table instead of the auto-generated column size (currently 100, in the future, we add maxLength field in the schema).

Testing done:
- updated unit tests
- manual test in Synapse of shrinking and growing columns - Made sure that even when the table is updated, it retained the previous max size in the columns and didn't delete any data.


alxdarksage:
Ok, looks good, merging

PR #14, opened 2016-05-10
support for start/endDateTime to enable hourly and on-demand exports

DwayneJengSage:
There are now two indices on the records table, one on uploadDate and one on studyId+uploadedOn. This enables us to use a range query on uploadedOn, which enable hourly exports and on-demand exports.

The one limitation is uploadedOn must be a range key, so we use the study ID as the hash key. This is fine, because uploadedOn will only be used on specific studies.

This also has the side effect of optimizing our studyWhitelist usage. Before, we'd get every record for a day and filter out the ones that don't match the study. Now, we can use the studyId-uploadedOn-index to only query for the records in our study.

Testing done:
- added unit tests
- manual tests (a) studyWhitelist w/ start/endDateTimes (b) studyWhitelist w/ date (c) all studies w/ date (normal use case)

Next steps:
- A mechanism so studies don't get exported by both the hourly schedule and the "default" schedule
- Configure a separate scheduler for hourly exports

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
@@ -3,38 +3,47 @@ package org.sagebionetworks.bridge.exporter.request;
 import java.util.Map;
 import java.util.Set;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.datatype.joda.deser.LocalDateDeserializer;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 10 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -109,12 +138,43 @@ public String getTag() {
...
+            stringBuilder.append("recordIdS3Override=");
+            stringBuilder.append(recordIdS3Override);
+        }
+
+        // Always include tag.
+        stringBuilder.append(", tag=");
+        stringBuilder.append(tag);
+
+        return stringBuilder.toString();
+    }
```
alxdarksage, line 169: is this just the appropriate toString() method for this object?

DwayneJengSage, line 169: Changed to toString()

PR #15, opened 2016-05-11
usesCustomExportSchedule flag

DwayneJengSage:
Update BridgeEX to parse the usesCustomExportSchedule flag and honor it.

We currently determine what is and isn't a custom export schedule based on the presence or absence of the study whitelist flag, which is needed to use the granular uploadedOn index for hourly exports. In the future, this may become more sophisticated, and we may need to shift to passing in an explicit customSchedule in the request object, but for now, I think this is a viable solution.

Testing done:
- mvn verify (findbugs, unit tests, test coverage)
- manual tests: (a) custom schedule false (b) custom schedule true w/o study whiteliest (c) custom schedule true w/ study whitelist

Next steps:
- set up hourly scheduler

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/StudyInfo.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/StudyInfo.java
@@ -29,8 +29,8 @@ public class StudyInfo {
     }
 
     /**
-     * True if BridgeEX should include it in the "default" nightly job (which is an export job without a study
-     * whitelist). False otherwise.
+     * False if BridgeEX should include it in the "default" nightly job (which is an export job without a study
+     * whitelist). True otherwise.
      */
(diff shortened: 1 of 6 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/dynamo/StudyInfo.java
```diff
@@ -26,10 +28,19 @@ public String getSynapseProjectId() {
         return synapseProjectId;
     }
 
+    /**
+     * True if BridgeEX should include it in the "default" nightly job (which is an export job without a study
```
liujoshua: Is this comment inverted?

DwayneJengSage: You're right, I got it backwards. Will fix.

PR #24, opened 2016-07-12
Listen to S3 file uploaded notifications and call uploadComplete on bridge

liujoshua:


src/main/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallback.java
```diff
@@ -0,0 +1,75 @@
...
+     * minutes. Package-scoped to enable unit tests to mock this out.
+     */
+    @Cacheable(lifetime = 5, unit = TimeUnit.MINUTES)
+    WorkerClient getWorkerClient() {
+        Session session = ClientProvider.signIn(credentials);
+        return session.getWorkerClient();
+    }
+
+    @Override
+    public void callback(String messageBody) throws Exception {
```
alxdarksage, line 56: If this throws an exception, shouldn't we swallow it but log it instead? Or do we know that when this throws, the encompassing framework handles logging, it doesn't kill the thread, etc.?

src/main/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallback.java
```diff
@@ -0,0 +1,75 @@
...
+
+        callback(notification);
+    }
+
+    // package-scoped to enable mocking/testing
+    void callback(S3EventNotification notification) {
+        notification.getRecords().stream().filter(record -> shouldProcessRecord(record)).forEach(record -> {
+            String uploadId = record.getS3().getObject().getKey();
+            LOG.info("Completing upload, id=" + uploadId);
+            getWorkerClient().completeUpload(uploadId);
```
alxdarksage, line 67: What if we flush Redis and you lose your session in the 5 minutes this is cached, what happens?

src/test/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallbackTest.java
```diff
@@ -0,0 +1,131 @@
...
+        when(record.getS3()).thenReturn(entity);
+
+        return record;
+    }
+
+    @Test
+    public void testShouldProcessRecords_S3Put() {
+        S3EventNotification.S3EventNotificationRecord s3Put = createMockRecord("aws:s3", "ObjectCreated:Put");
+
+        assert callback.shouldProcessRecord(s3Put);
```
alxdarksage: I would prefer assertTrue()

src/main/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallback.java
liujoshua, line 56: By throwing an exception, we are allowing SQS to redeliver this message.

When callback() throws an Exception, PollSqsWorker's run() method swallows it and logs the message (and skips the part where it removes the message from the queue)

alxdarksage, line 56: Okay. Does this create a loop if the error is on the Bridge side?

src/main/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallback.java
liujoshua, line 67: Looks like the worker client will remain unauthorized. The message should get redelivered, which isn't ideal but we shouldn't lose a message. I'll think about better ways to authenticate

src/test/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallbackTest.java
liujoshua: Fixed

src/main/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallback.java
alxdarksage, line 67: That's probably fine, though if the thing stays cached for 5 minutes, that could be a lot of failed messages.

src/main/java/org/sagebionetworks/bridge/exporter/notification/S3EventNotificationCallback.java
liujoshua, line 56: In a sense. There are settings in SQS for how long messages stay invisible before qualifying for redelivery, how many times a message can be redelivered, and a maximum retention period before a message gets deleted

PR #27, opened 2016-07-25
pull in connection leak fixes for bridge-base and SynapseJavaClient

DwayneJengSage:
We have a semi-regular problem with connection timeouts causing BridgeEX to fail catastrophically. This pulls in changes for bridge-base and SynapseJavaClient which should hopefully fix some, if not all, connection issues.

Testing done:
- manual tested a redrive scenario, which pulls a list of record IDs from an S3 file (the main way this hits that bridge-base code path

This is dependent on https://github.com/Sage-Bionetworks/bridge-base/pull/25

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <dependency>
             <groupId>org.sagebionetworks</groupId>
             <artifactId>bridge-base</artifactId>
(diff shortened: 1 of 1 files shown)


liujoshua:
Looks good, merge when ready

PR #28, opened 2016-07-27
Check Synapse writable before export job

DwayneJengSage:
This is the code change for the first half of https://sagebionetworks.jira.com/browse/BRIDGE-1401

This change adds a check for Synapse status at the start of each Export run. If Synapse is not writable, we throw a SynapseUnavailableException, which allows SQS to re-cycle the request and try later (currently configured at 2 hours; 2 hours chosen because in some unusual cases, Bridge Exporter may take up to ~1.5 hours to complete).

Testing done:
- manual test against Synapse staging in read-only
- manual test for normal case


alxdarksage:
reviewer=alxdark

PR #29, opened 2016-08-18
fix Health Data Handler holding onto schemas forever

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-1454

Health Data Handler stores a copy of the schema and has no mechanism to update it. This effectively caches schemas forever. This is bad, because it prevents mutable schemas from being mutable.

The fix is to have Health Data Handler store a reference to the schema key, and then call Bridge for the schema (which is cached for 5 minutes via BridgeHelper).

Testing done:
- manually tested running the exporter, updating the schema, and running the exporter again to verify schema changes were picked up


liujoshua:
Looks good

DwayneJengSage:
Thanks! Please merge whenever you're ready.

PR #31, opened 2016-09-02
Fix redrive bug; improve Bridge Server sessions

DwayneJengSage:
Three fixes:
1. Fix session handling in BridgeHelper. This new code caches sessions "forever", but has handling to detect when sessions expire and refresh the session.
2. Added AmazonClientException to the catch block when redrives. This allows us to fail gracefully if the S3 upload fails or the SQS send fails.
3. Max delay for SQS is 900 seconds (15 min).

Testing done:
- manual tests

Manual tests:
- normal case
- manually expire session, verified BridgeEX refreshes sessions in logs
- inject error, block upload to S3 (with permissions), verify BridgeEX fails gracefully
- inject error, use test delay of 30 seconds, verify BridgeEX redrives successfully
- inject error, use real delay of 900 seconds, verify the message is successfully sent to SQS


liujoshua:
Looks good!

PR #32, opened 2016-09-20
improved session handling for Upload Auto-Complete

DwayneJengSage:
Currently, Upload Auto-Complete caches the Bridge session for 5 minutes. This means, when Bridge session expires, for up to 5 minutes, we're unable to auto-complete uploads.

BridgeHelper in the Bridge-EX project already has better session management, which catches the 401, signs in, and retries the call. This change updates Upload Auto-Complete to use BridgeHelper instead of managing sessions directly. It also updates error handling to propagate 401s and 403s, which indicate a problem with the client rather than the request.

Testing done:
- manual tests to verify we automatically refresh the session
- rigged up local Bridge server to verify new error handling logic
- updated unit tests
- mvn verify (unit tests, jacoco test coverage, findbugs)

See https://sagebionetworks.jira.com/browse/BRIDGE-1485

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelper.java
@@ -34,6 +34,21 @@ public class BridgeHelper {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


liujoshua (approved):

liujoshua:
Thanks for fixing this

PR #33, opened 2016-10-13
update bridgeExporter to hook up new api to update record exporter status

zaclinsage:
update bridgeHelper to call new api to update record export status
    keep track of record ids in tsv file and update record exporter status by calling bridgehelper new method
    also add corresponding unit tests for above codes.


pom.xml
```diff
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>org.sagebionetworks.bridge</groupId>
             <artifactId>java-sdk</artifactId>
-            <version>0.8.21</version>
+            <version>0.10.7</version>
```
alxdarksage: I think this is now bumped to 0.11.1

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
```diff
@@ -82,6 +88,11 @@ protected String getDdbTableKeyValue() {
         return getSynapseTableColumnListCached(schemaFieldDefList);
     }
 
+    @Override
+    public void setBridgeHelper(BridgeHelper bridgeHelper) {
+        this.bridgeHelper = bridgeHelper;
+    }
```
alxdarksage: Is this being injected by Spring? If so you can remove initialization to an instance. Is this not being injected? Then can you make a static final property of this class?

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
```diff
@@ -217,6 +228,20 @@ protected void setTsvInfoForTask(ExportTask task, TsvInfo tsvInfo) {
...
 
+    /**
+     * post process tsv to call update records' exporter status as SUCCEEDED
+     * @param tsvInfo
+     */
+    @Override
+    protected void postProcessTsv(TsvInfo tsvInfo) throws BridgeExporterException {
+        List<String> recordIds = tsvInfo.getRecordIds();
+        if (recordIds == null) {
+            throw new BridgeExporterException("No record id in this tsv file.");
```
alxdarksage: Are we sure this is an error condition? For example if no uploads occur in a study for an entire day or whatever the export period is, would it throw an exception where none is warranted?

src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
```diff
@@ -37,6 +38,7 @@ private TsvInfo() {
         this.file = null;
         this.writer = null;
         this.initError = true;
+        this.recordIds = null;
```
alxdarksage: I guess I would just this.recordIds = new ArrayList<>() right here, instead of requiring callers to pass it in.

src/test/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandlerTest.java
```diff
@@ -1,23 +1,13 @@
...
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertNull;
-import static org.testng.Assert.assertTrue;
-import static org.testng.Assert.fail;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
+import static org.testng.Assert.*;
```
alxdarksage: We usually import these individually. It's trivial, but I wouldn't break with this as the next checkin, there'll be a diff putting them all back again.

src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -392,4 +398,22 @@ public void uploadToSynapseForTask(ExportTask task) throws BridgeExporterExcepti
...
+
+    };
+
+    /**
+     * dummy method to implement by HealthDataExportHandler for testing
+     * @param bridgeHelper
+     */
+    protected void setBridgeHelper(BridgeHelper bridgeHelper) {
+
+    }
```
alxdarksage: I don't love these... do both ExportHandlers have to have all the same methods? Can these be removed somehow? These aren't declared in ExportHandler so it seems like it should be possible to remove them.

alxdarksage (changes requested):
Some comments.

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
DwayneJengSage: To add to this, HealthDataExportHandler already has the BridgeHelper through getManager().getBridgeHelper().

Also, don't override dependency injection setters. This leads to bad things happening.

src/test/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandlerTest.java
zaclinsage: oh i diddn't do that. the ide seems like to add it if i rely on all assert commands in this lib

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
zaclinsage: well yes i think the null is not permitted here. I changed the initialization to create a new arraylist instead of a null

src/test/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandlerTest.java
zaclinsage: i fixed it by changing ide settings

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
zaclinsage: delete my setter by using manager's one

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
```diff
@@ -45,6 +50,7 @@
     private static final long TIME_ZONE_FIELD_LENGTH = 5;
     private static final DateTimeFormatter TIME_ZONE_FORMATTER = DateTimeFormat.forPattern("Z");
     private static final String TIME_ZONE_UTC_STRING = "+0000";
+    private BridgeHelper bridgeHelper = new BridgeHelper();
```
DwayneJengSage: Remove unused bridgeHelper.

DwayneJengSage:
Also, delete dump.rdb. This file is created by redis-server in the local directory and shouldn't be included in our codebase.

src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
```diff
@@ -49,11 +51,12 @@ private TsvInfo() {
      * @param writer
      *         writer for the TSV file
      */
-    public TsvInfo(List<String> columnNameList, File file, PrintWriter writer) {
+    public TsvInfo(List<String> columnNameList, File file, PrintWriter writer, List<String> recordIds) {
         this.columnNameList = columnNameList;
         this.file = file;
         this.writer = writer;
         this.initError = false;
+        this.recordIds = recordIds;
```
DwayneJengSage: recordIds is always an empty list, so instead of having the constructors create it, we should initialize it in the declaration above.

src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
```diff
@@ -21,6 +21,7 @@
     private final File file;
     private final PrintWriter writer;
     private final boolean initError;
+    private List<String> recordIds;
```
DwayneJengSage: Mark this as final. This ensures we aren't overwriting this somewhere.

src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
```diff
@@ -87,6 +90,18 @@ public int getLineCount() {
...
+     * helper method to add a record id into the list
+     * @param recordId
+     */
+    public void addRecordId(String recordId) {
+        this.recordIds.add(recordId);
+    }
+
+    public List<String> getRecordIds() {
+        return this.recordIds;
+    }
```
DwayneJengSage, line 101: For safety, I would wrap this in an ImmutableList.copyOf(this.recordIds). This ensures that whoever is calling this doesn't end up accidentally modifying it.

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
DwayneJengSage: Since we're now always initializing recordIds to an empty list, this will never be null.

src/main/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandler.java
```diff
@@ -16,6 +16,11 @@
 import org.joda.time.DateTime;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
+import org.sagebionetworks.bridge.exporter.config.SpringConfig;
+import org.sagebionetworks.bridge.exporter.exceptions.BridgeExporterException;
+import org.sagebionetworks.bridge.exporter.helper.BridgeHelper;
+import org.sagebionetworks.bridge.sdk.models.accounts.SignInCredentials;
+import org.sagebionetworks.bridge.sdk.models.healthData.RecordExportStatusRequest;
```
DwayneJengSage, line 20: Minor: Remove unused imports "SpringConfig" and "SignInCredentials". I believe by default, IntelliJ marks these as warnings. (Although they might have changed this.)

src/main/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelper.java
```diff
@@ -49,6 +53,23 @@ public void completeUpload(String uploadId) {
     }
 
     /**
+     * Helper method to update export status in records
+     */
+    public void updateRecordExporterStatus(List<String> recordIds, RecordExportStatusRequest.ExporterStatus status) {
+        // first create a credentials for this session
+        SpringConfig config = new SpringConfig();
+        SignInCredentials credentials = config.bridgeWorkerCredentials();
+        setCredentials(credentials);
```
DwayneJengSage: Don't do this. Spring is supposed to take care of credentials.

src/main/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelper.java
```diff
@@ -49,6 +53,23 @@ public void completeUpload(String uploadId) {
...
+
+        // then update status
+        RecordExportStatusRequest request = new RecordExportStatusRequest(recordIds, status);
+        sessionHelper(() -> {
+            session.getWorkerClient().updateRecordExporterStatus(request);
+            return null;
+        });
```
DwayneJengSage, line 76: List of record IDs may be larger than your batch size. You'll need to break down the list into batches and call Bridge in a loop, sleeping between each batch. (Most likely sleeping for 1 second between each batch.)

src/test/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandlerTest.java
```diff
@@ -401,6 +414,12 @@ public void healthDataExportHandlerTest() throws Exception {
         // validate tsvInfo
         TsvInfo tsvInfo = handler.getTsvInfoForTask(task);
         assertEquals(tsvInfo.getLineCount(), 1);
+        assertNotNull(tsvInfo.getRecordIds());
+        List<String> recordIds = tsvInfo.getRecordIds();
+        assertEquals(recordIds.size(), 1);
+        for (String recordId : recordIds) {
+            assertEquals(recordId, DUMMY_RECORD_ID);
+        }
```
DwayneJengSage, line 422: We should also verify that BridgeHelper.recordExportStatus() was called.

src/test/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelperTest.java
```diff
@@ -64,6 +71,9 @@ private static BridgeHelper setupBridgeHelperWithSession(Session session) {
         return bridgeHelper;
     }
 
+    ArgumentCaptor<RecordExportStatusRequest> requestArgumentCaptor = ArgumentCaptor.forClass(RecordExportStatusRequest.class);
```
DwayneJengSage: It's not clear to me that TestNG creates a new test class instance before each test. You should initialize this in a @BeforeMethod instead.

src/test/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelperTest.java
```diff
@@ -78,6 +88,23 @@ public void completeUpload() {
...
+        Session mockSession = mock(Session.class);
+        when(mockSession.getWorkerClient()).thenReturn(mockWorkerClient);
+        BridgeHelper bridgeHelper = setupBridgeHelperWithSession(mockSession);
+
+        // execute and verify
+        bridgeHelper.updateRecordExporterStatus(TEST_RECORD_IDS, TEST_STATUS);
+        verify(mockWorkerClient).updateRecordExporterStatus(anyVararg());
+        verify(mockWorkerClient).updateRecordExporterStatus(requestArgumentCaptor.capture());
+        RecordExportStatusRequest request = requestArgumentCaptor.getValue();
+        assertEquals(request.getRecordIds().get(0), TEST_RECORD_IDS.get(0));
```
DwayneJengSage: You can actually just call equals() on lists.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelper.java
```diff
@@ -62,11 +64,21 @@ public void updateRecordExporterStatus(List<String> recordIds, RecordExportStatu
...
+        batches.forEach(batch-> {
+            RecordExportStatusRequest request = new RecordExportStatusRequest(batch, status);
+            sessionHelper(() -> {
+                session.getWorkerClient().updateRecordExporterStatus(request);
+                return null;
+            });
+            try {
+                TimeUnit.SECONDS.sleep(1);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
DwayneJengSage: Minor: Instead of e.printStackTrace(), use LOG.warn(). This will ensure our output goes to the logs.

src/test/java/org/sagebionetworks/bridge/exporter/helper/BridgeHelperTest.java
```diff
@@ -78,6 +93,23 @@ public void completeUpload() {
...
+        Session mockSession = mock(Session.class);
+        when(mockSession.getWorkerClient()).thenReturn(mockWorkerClient);
+        BridgeHelper bridgeHelper = setupBridgeHelperWithSession(mockSession);
+
+        // execute and verify
+        bridgeHelper.updateRecordExporterStatus(TEST_RECORD_IDS, TEST_STATUS);
+        verify(mockWorkerClient).updateRecordExporterStatus(anyVararg());
+        verify(mockWorkerClient).updateRecordExporterStatus(requestArgumentCaptor.capture());
+        RecordExportStatusRequest request = requestArgumentCaptor.getValue();
+        request.getRecordIds().equals(TEST_RECORD_IDS);
```
DwayneJengSage: You need an assert here.

DwayneJengSage (changes requested):

PR #34, opened 2016-10-26
switch from SynapseClient.createFileHandle() to SynapseClient.multipartUpload()

DwayneJengSage:
BridgeEX has frequent connection issues uploading file handles to Synapse.

We determined that this is because BridgeEX uses the old deprecated SynapseClient.createFileHandle(), which may have connection leak issues. We switched to the new SynapseClient.multipartUpload(), which is much more robust.

Testing done:
- manual tests, include tests with an attachment (file handle)


alxdarksage:
Looks good to me.

PR #35, opened 2016-12-14
migrate to new Bridge REST Client

DwayneJengSage:
This migrates to the new Bridge REST Client, which ensures that we're using the most up-to-date APIs and library versions.

Testing done:
* manual tests

Manual Testing done:
* export, using schema upload-v2-manual-test to test a variety of fields
* verified recordExportStatus
* upload autocomplete
* refresh sessions (both export and upload autocomplete)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,11 @@
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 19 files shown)


alxdarksage:
Looks good to merge. I'm curious if you're working around the fact that the client doesn't re-authenticate when it loses a session?

PR #36, opened 2016-12-15
improve redrive logic

DwayneJengSage:
This change improves BridgeEX's redrive logic. First, it differentiates between retryable and non-retryable errors, so BridgeEX won't be stuck in a loop retrying exports that are doomed to failure.

Non-retryable errors include:
* 4XX errors from Bridge Server
* JSON parse exceptions
* mismatched schema errors

Second, when a TSV fails to initiate, BridgeEX now only redrives the table and not the records, to prevent a "double redrive". (Note this doesn't fix the general case, where a record fails and its table also fails. These tend to be fairly rare. The fix would also be more complicated, and likely outside the scope of a short-term DevOps fix.)

These fixes are in response to DevOps issues resulting from the mythoughts schema incident and the recordExportStatus incident. See also https://sagebionetworks.jira.com/browse/BRIDGE-1633

Testing done:
* manual tests

Manual test cases:
* successful export
* modified schema error (mythoughts schema case)
* bad BridgePF URL (record export status case)
* inject errors (retryable/non-retryble x record/table)

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManager.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManager.java
@@ -20,6 +20,8 @@ import com.amazonaws.services.dynamodbv2.document.Table;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.base.Stopwatch;
+import com.google.gson.JsonParseException;
+import com.google.gson.stream.MalformedJsonException;
 import org.apache.commons.lang.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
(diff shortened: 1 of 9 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManager.java
```diff
@@ -627,4 +644,34 @@ public void endOfStream(ExportTask task) throws RestartBridgeExporterException {
...
+            if (statusCode >= 400 && statusCode < 500) {
+                return false;
+            }
+        } else if (t instanceof JsonProcessingException || t instanceof JsonParseException ||
+                t instanceof MalformedJsonException) {
+            // JSON parse exceptions are generally also not retryable. Note that BridgeEX uses Jackson, but
+            // BridgeJavaSDK uses GSON, so we need to handle both.
+            // Jackson uses JsonProcessingException as their base exception.
+            // GSON uses JsonParseException and MalformedJsonException.
+            return false;
```
alxdarksage, line 668: Should we file a JIRA to hide dependent library exceptions like these and rethrow them as BridgeSDKExceptions? It seems unfortunate you have to know about the rest-client's serialization implementation. We'd have to figure out if we can intercept this in the generated code, I think.

alxdarksage (commented):
You don't have to do this, it's just an idea: the non-retryability of these exceptions could be indicated with an annotation, and business-specific exceptions like BridgeExporterJsonParsingException, BridgeExporterModfiedSchemaException, etc. could be used for the various cases that should not be retried. Later it might be clearer.

PR #38, opened 2016-12-24
Rate Limiting and Metrics for Synapse requests

DwayneJengSage:
A few changes in response to Synapse Throttling:
* Add RateLimiter to SynapseHelper.
* Default config is now 4 threads.
* Logging now indicates whether request succeeded or failed.
* Count the number of file handles.

This also does a few test refactors.

Testing done:
* manually tested successful export with attachment
* injected error, manually verified log reports error

### DIFF
--- a/src/main/resources/BridgeExporter.conf
+++ b/src/main/resources/BridgeExporter.conf
@@ -16,7 +16,8 @@ record.loop.delay.millis=30
 record.loop.progress.report.period=1000
 synapse.async.interval.millis = 1000
 synapse.async.timeout.loops = 300
-threadpool.worker.count=16
+synapse.rate.limit.per.second = 10
+threadpool.worker.count=4
 time.zone.name=America/Los_Angeles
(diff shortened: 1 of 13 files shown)


src/main/resources/BridgeExporter.conf
```diff
@@ -16,7 +16,8 @@ record.loop.delay.millis=30
 record.loop.progress.report.period=1000
 synapse.async.interval.millis = 1000
 synapse.async.timeout.loops = 300
-threadpool.worker.count=16
+synapse.rate.limit.per.second = 10
```
liujoshua, line 19: Does the 10/s need to be split among the four worker threads?

liujoshua (approved):

src/main/resources/BridgeExporter.conf
DwayneJengSage: SynapseHelper is a singleton, so it limits the rate across all 4 threads.

PR #39, opened 2017-01-10
add userSharingScope field when exporting to Synapse table

zaclinsage:


zaclinsage:
test done: 
1. all unit tests in package;
2. manual test: upload test data through Bridge Server and send SQS msg to exporter; then verify if new column was added;
3. mvn clean test / mvn clean verify;
4. add more tests: unit tests for TransferMethod, serialize/de-serialize test for ColumnDefinition, test cases for no common column list table

src/test/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandlerTest.java
```diff
@@ -42,7 +47,14 @@
 import org.sagebionetworks.bridge.rest.model.UploadFieldDefinition;
 import org.sagebionetworks.bridge.rest.model.UploadFieldType;
 
-public class HealthDataExportHandlerTest {
+import javax.annotation.Resource;
+
+@ContextConfiguration("classpath:test-context.xml")
+public class HealthDataExportHandlerTest extends AbstractTestNGSpringContextTests {
```
DwayneJengSage: Strongly recommend against introducing Spring into unit tests, especially mock tests that don't hit real dependencies. It adds unnecessary overhead and complexity to our tests, makes them more difficult to manage, and you can do this testing without using Spring anyway. This also means you're essentially testing the same thing in 4 different places. Also, if we ever change the configs, we'll need to change the tests in 4 different places, which is exactly what we wanted to avoid doing with this change.

Here and in SynapseExportHandlerNewTableTest, SynapseExportHandlerTest, and SynapseExportHandlerUpdateTableTest.

pom.xml
```diff
@@ -97,6 +97,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-logging</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <version>4.0.5.RELEASE</version>
+        </dependency>
```
DwayneJengSage: We don't want to introduce Spring into our unit tests. Please remove.

src/main/java/org/sagebionetworks/bridge/exporter/config/SpringConfig.java
```diff
@@ -194,4 +200,15 @@ public SynapseClient synapseClient() {
     public ExecutorService workerExecutorService() {
         return Executors.newFixedThreadPool(bridgeConfig().getInt("threadpool.worker.count"));
     }
+
+    @Bean(name = "synapseColumnDefinitions")
+    public List<ColumnDefinition> synapseColumnDefinitions() throws IOException {
+        final ClassLoader classLoader = getClass().getClassLoader();
+        final File file = new File(classLoader.getResource("ColumnDefinition.json").getFile());
+        final ObjectMapper mapper = new ObjectMapper();
```
DwayneJengSage: Minor: Instead of instantiating your own ObjectMapper, you can use DefaultObjectMapper.INSTANCE from bridge-base.

src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -47,8 +50,14 @@
     private static final Logger LOG = LoggerFactory.getLogger(SynapseExportHandler.class);
 
     // Package-scoped to be available to unit tests.
-    static final List<ColumnModel> COMMON_COLUMN_LIST;
-    static {
+    List<ColumnModel> COMMON_COLUMN_LIST;
```
DwayneJengSage: Minor: Should be commonColumnList, not COMMON_COLUMN_LIST.

src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -47,8 +50,14 @@
...
+
+    private List<ColumnDefinition> COLUMN_DEFINITION;
```
DwayneJengSage: Similarly here.

src/main/java/org/sagebionetworks/bridge/exporter/config/SpringConfig.java
```diff
@@ -194,4 +200,15 @@ public SynapseClient synapseClient() {
...
+
+        List<ColumnDefinition> value = mapper.readValue(file, new TypeReference<List<ColumnDefinition>>(){});
+
+        return value;
```
DwayneJengSage: Recommend copying the result into an ImmutableList. This prevents consuming classes from accidentally modifying the bean.

Also, our manual test case should have multiple tables to make sure we're handling this correctly.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/TransferMethod.java
```diff
@@ -0,0 +1,52 @@
...
+        public String transfer(String ddbName, Item record, ExportTask task) {
+            return String.valueOf(record.getLong(ddbName));
+        }
+    },
+    EXPORTERDATE {
+        @Override
+        public String transfer(String ddbName, Item record, ExportTask task) {
+            return task.getExporterDate().toString();
+        }
+    };
```
DwayneJengSage, line 57: I don't think ExporterDate is useful. This is specific to one column, and the ExportTask is just exporter state and is unlikely to have new fields that need exporting. We should remove this.

src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -57,36 +66,6 @@
...
-        dataGroupsColumn.setColumnType(ColumnType.STRING);
-        dataGroupsColumn.setMaximumSize(100L);
-        columnListBuilder.add(dataGroupsColumn);
-
-        // NOTE: ColumnType.DATE is actually a timestamp. There is no calendar date type.
-        ColumnModel uploadDateColumn = new ColumnModel();
-        uploadDateColumn.setName("uploadDate");
-        uploadDateColumn.setColumnType(ColumnType.STRING);
-        uploadDateColumn.setMaximumSize(10L);
-        columnListBuilder.add(uploadDateColumn);
```
DwayneJengSage, line 83: Similarly, don't delete uploadDateColumn here.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,164 @@
...
+    private String name;
+    private ColumnType columnType;
+    private Long maximumSize;
+    private TransferMethod transferMethod;
+    private String ddbName;
+    private Boolean sanitize;
+
+    public ColumnDefinition() {
+
+    }
```
DwayneJengSage: Minor: Don't need to explicitly create a constructor if it has no args and no body.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,164 @@
...
+
+import org.sagebionetworks.repo.model.table.ColumnType;
+
+public class ColumnDefinition  {
+    private String name;
+    private ColumnType columnType;
+    private Long maximumSize;
+    private TransferMethod transferMethod;
+    private String ddbName;
+    private Boolean sanitize;
```
DwayneJengSage: Minor: For sanitize, since we treat the null case the same as the false case, we could just use a lower-case boolean and default to false. This will save us null checks elsewhere.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,164 @@
...
+        int result1 = 31 * result + (this.name == null?0:this.name.hashCode());
+        result1 = 31 * result1 + (this.columnType == null?0:this.columnType.hashCode());
+        result1 = 31 * result1 + (this.maximumSize == null?0:this.maximumSize.hashCode());
+        result1 = 31 * result1 + (this.transferMethod == null?0:this.transferMethod.hashCode());
+        result1 = 31 * result1 + (this.ddbName == null?0:this.ddbName.hashCode());
+        result1 = 31 * result1 + (this.sanitize == null?0:this.sanitize.hashCode());
+        return result1;
+    }
+
+    public boolean equals(Object obj) {
```
DwayneJengSage: Do we use .equals() on ColumnDefinitions anywhere, or use it in a hash map? If not, then we don't need hashCode() and equals().

If we *do* need hashCode() and equals(), we need to test it. I recommend using EqualsVerifier. See BridgeExporterRequestTest for example.

Also, keep in mind that mutable objects don't work as hash keys. If you need to use this as a hash key to a hash map, strongly recommend making this class immutable and use the builder pattern instead.

Lastly, you can use Eclipse or IntellJ to auto-generate hashCode() and equals() to make them less wordy. Recommend using either Guava Objects or Java 8 Objects to generate these.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,164 @@
...
+                }
+            } else if(!this.sanitize.equals(other.sanitize)) {
+                return false;
+            }
+
+            return true;
+        }
+    }
+
+    public String toString() {
```
DwayneJengSage: Similarly, this can be auto-generated. Also, recommend having a test with all null values to verify that this doesn't throw an NPE if we call toString().

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,164 @@
+package org.sagebionetworks.bridge.exporter.synapse;
+
+import org.sagebionetworks.repo.model.table.ColumnType;
+
+public class ColumnDefinition  {
```
DwayneJengSage, line 5: We need a test that serializes and deserializes to/from JSON to make sure this works the way we expect.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/TransferMethod.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Enum class representing several different transfer method from record value in ddb table to synapse row value.
+ * Also specifying corresponding transfer procedure.
+ */
+public enum TransferMethod {
```
DwayneJengSage, line 16: Needs unit tests.

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -1,15 +1,20 @@
 package org.sagebionetworks.bridge.exporter.util;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
DwayneJengSage: Minor: Our convention is to show all our imports instead of collapsing them with *

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -1,15 +1,20 @@
...
 
 import com.amazonaws.services.dynamodbv2.document.Item;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSetMultimap;
 import org.jsoup.Jsoup;
 import org.jsoup.safety.Whitelist;
+import org.omg.PortableInterceptor.SYSTEM_EXCEPTION;
```
DwayneJengSage: What is SYSTEM_EXCEPTION?

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -21,6 +26,8 @@
 public class BridgeExporterUtil {
     private static final Logger LOG = LoggerFactory.getLogger(BridgeExporterUtil.class);
 
+    public static final Joiner STRING_SET_JOINER = Joiner.on(',').useForNull("");
```
DwayneJengSage, line 31: Minor: If we're moving DATA_GROUP_JOINER here, we should delete the original (now unused) one from SynapseExportHandler.

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -19,6 +33,12 @@
 import org.sagebionetworks.bridge.schema.UploadSchemaKey;
 
 public class BridgeExporterUtilTest {
+    private List<ColumnModel> TEST_COLUMN_MODELS;
+    private List<ColumnDefinition> TEST_COLUMN_DEFINITIONS;
```
DwayneJengSage: Minor: non-static member variables should be camel-cased (testColumnModles), not all-caps (TEST_COLUMN_MODELS).

Also, it looks like these aren't being used.

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +222,132 @@ public void shouldConvertFreeformTextToAttachment() {
                 .withRevision(1).build();
         assertFalse(BridgeExporterUtil.shouldConvertFreeformTextToAttachment(other, "other"));
     }
+
+    @Test
+    public void canConvertToColumnList() {
+        final String TEST_COLUMN_MODEL_NAME_1 = "test_column_model_1";
+        final String TEST_COLUMN_MODEL_NAME_2 = "test_column_model_2";
```
DwayneJengSage: Local variables should also be camel-cased.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,164 @@
...
+    private String name;
+    private ColumnType columnType;
+    private Long maximumSize;
+    private TransferMethod transferMethod;
```
DwayneJengSage, line 8: Can we combine TransferMethod and ColumnType? It looks like these map one-to-one, and I can't think of any case where they'd be different. The simplest thing to do is for TransferMethod to have a getColumnType() method that returns the appropriate ColumnType.

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -185,4 +192,44 @@ public static boolean shouldConvertFreeformTextToAttachment(UploadSchemaKey sche
...
+
+        return columnListBuilder.build();
+    }
+
+    public static Map<String, String> getRowValuesFromRecordBasedOnColumnDefinition(final Map<String, String> rowMap, final Item record, final List<ColumnDefinition> columnDefinitions, final String recordId, final ExportTask task) {
+        Map<String, String> retMap = new HashMap<>();
+        retMap.putAll(rowMap);
+
+        for (ColumnDefinition columnDefinition : columnDefinitions) {
+            final String ddbName = columnDefinition.getDdbName();
```
DwayneJengSage: Would it be worth adding a feature so that if ddbName is null, we use name for the ddbName? This way, we don't always have to type every name twice in the JSON configs.

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -185,4 +192,44 @@ public static boolean shouldConvertFreeformTextToAttachment(UploadSchemaKey sche
...
+            columnModel.setMaximumSize(columnDefinition.getMaximumSize());
+            columnListBuilder.add(columnModel);
+        }
+
+        return columnListBuilder.build();
+    }
+
+    public static Map<String, String> getRowValuesFromRecordBasedOnColumnDefinition(final Map<String, String> rowMap, final Item record, final List<ColumnDefinition> columnDefinitions, final String recordId, final ExportTask task) {
+        Map<String, String> retMap = new HashMap<>();
+        retMap.putAll(rowMap);
```
DwayneJengSage: Instead of intsantiating a map, copying the contents, and returning the new map, maybe we can have this method update the map in-place and have a void return type. This way, we don't have to instantiate more maps then we need, and it makes the code in SynapseExportHelper easier to reason about. (Right now, we take a map, copy it, and then add all those values back to itself.)

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +222,132 @@ public void shouldConvertFreeformTextToAttachment() {
...
+        assertEquals(BridgeExporterUtil.convertToColumnList(TEST_COLUMN_DEFINITIONS), TEST_COLUMN_LIST);
+    }
+
+    @Test
+    public void canGetRowValuesFromRecordBasedOnColumnDefinition() {
+        final String TEST_STRING_NAME = "test_string";
+        final String TEST_STRING_SET_NAME = "test_string_set";
+        final String TEST_DATE_NAME = "test_date";
+        final String TEST_EXPORTER_DATE_NAME = "test_exporter_date";
+        final String TEST_SANITIZE = "test_sanitize";
```
DwayneJengSage: Similarly here.

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +222,132 @@ public void shouldConvertFreeformTextToAttachment() {
...
+        ExportTask mockTask = mock(ExportTask.class);
+        when(mockTask.getExporterDate()).thenReturn(TEST_LOCAL_DATE);
+
+        // create mock record
+        Item testRecord = new Item();
+        testRecord.withString(TEST_STRING_NAME, "test_string_value");
+        testRecord.withStringSet(TEST_STRING_SET_NAME, new String[]{"test_string_set_value_1", "test_string_set_value_2"});
+        testRecord.withLong(TEST_DATE_NAME, 1484181511);
+        testRecord.withString(TEST_EXPORTER_DATE_NAME, "test_exporter_date_value");
+        testRecord.with(TEST_SANITIZE, "test_sanitize_value");
```
DwayneJengSage: The test_sanitize_value should have some HTML in it or something to verify we're properly calling the sanitization code.

src/main/resources/ColumnDefinition.json
```diff
@@ -0,0 +1,44 @@
...
+    "transferMethod": "STRINGSET",
+    "ddbName": "userDataGroups"
+  },
+  {
+    "name": "uploadDate",
+    "columnType": "STRING",
+    "maximumSize": "10",
+    "transferMethod": "EXPORTERDATE",
+    "ddbName": "uploadDate"
+  },
```
DwayneJengSage: Similarly, remove uploadDate from here.

src/main/resources/test-context.xml
```diff
@@ -0,0 +1,8 @@
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns:context="http://www.springframework.org/schema/context"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans
+        http://www.springframework.org/schema/beans/spring-beans.xsd http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context.xsd">
+    <context:annotation-config/>
+    <bean class="org.sagebionetworks.bridge.exporter.config.SpringConfig"/>
+</beans>
```
DwayneJengSage: Similarly, remove this.

DwayneJengSage (changes requested):
You should also add a test to SynapseExportHandlerUpdateTableTest with a table that doesn't have the common columns (or only has some of them) and test that your code successfully adds them to the mock TSV.

src/main/java/org/sagebionetworks/bridge/exporter/config/SpringConfig.java
```diff
@@ -194,4 +202,17 @@ public SynapseClient synapseClient() {
...
+
+    @Bean(name = "synapseColumnDefinitions")
+    public List<ColumnDefinition> synapseColumnDefinitions() throws IOException {
+        final ClassLoader classLoader = getClass().getClassLoader();
+        final File file = new File(classLoader.getResource("ColumnDefinition.json").getFile());
+        final ObjectMapper mapper = DefaultObjectMapper.INSTANCE;
+
+        List<ColumnDefinition> value = mapper.readValue(file, new TypeReference<List<ColumnDefinition>>(){});
+        ImmutableList.Builder<ColumnDefinition> columnDefinitionsBuilder = ImmutableList.builder();
+        columnDefinitionsBuilder.addAll(value);
```
DwayneJengSage: Minor: You can use ImmutableList.copyOf(value).

src/main/java/org/sagebionetworks/bridge/exporter/synapse/ColumnDefinition.java
```diff
@@ -0,0 +1,61 @@
+package org.sagebionetworks.bridge.exporter.synapse;
+
+import org.sagebionetworks.repo.model.table.ColumnType;
+
+public class ColumnDefinition  {
+    private String name;
+    private ColumnType columnType;
```
DwayneJengSage: Please remove columnType, as this no longer does anything.

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +217,123 @@ public void shouldConvertFreeformTextToAttachment() {
...
+        testDefinition2.setMaximumSize(48L);
+        columnDefinitionBuilder.add(testDefinition2);
+
+        testColumnDefinitions = columnDefinitionBuilder.build();
+
+        assertEquals(BridgeExporterUtil.convertToColumnList(testColumnDefinitions), testColumnList);
+    }
+
+    @Test
+    public void canGetRowValuesFromRecordBasedOnColumnDefinition() {
```
DwayneJengSage: We need a test case with name but not ddbName, to make sure we're testing that use case.

src/main/resources/ColumnDefinition.json
```diff
@@ -0,0 +1,34 @@
+[
+  {
+    "name": "healthCode",
+    "columnType": "STRING",
```
DwayneJengSage: Similarly, remove columnType from this file.

src/main/java/org/sagebionetworks/bridge/exporter/synapse/TransferMethod.java
```diff
@@ -0,0 +1,66 @@
+package org.sagebionetworks.bridge.exporter.synapse;
+
+import com.amazonaws.services.dynamodbv2.document.Item;
+import com.amazonaws.services.s3.transfer.Transfer;
```
DwayneJengSage: Minor: s3.transfer.Transfer is unused.

src/test/java/org/sagebionetworks/bridge/exporter/handler/HealthDataExportHandlerTest.java
```diff
@@ -43,6 +44,15 @@
 import org.sagebionetworks.bridge.rest.model.UploadFieldType;
 
 public class HealthDataExportHandlerTest {
+    private static List<ColumnModel> MOCK_COLUMN_LIST;
+
+    private static List<ColumnDefinition> MOCK_COLUMN_DEFINITION;
```
DwayneJengSage: Mark these as final. Same with the other 3 test files.

src/test/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandlerTest.java
```diff
@@ -461,13 +476,95 @@ public static void validateTsvHeaders(String line, String... extraColumnNameVara
...
+    }
+
+    public static List<ColumnDefinition> createTestSynapseColumnDefinitions() {
+        // setup column definitions
+        ImmutableList.Builder<ColumnDefinition> columnDefinitionsBuilder = ImmutableList.builder();
+
+        ColumnDefinition healthCodeDefinition = new ColumnDefinition();
+        healthCodeDefinition.setName("healthCode");
+        healthCodeDefinition.setDdbName("healthCode");
+        healthCodeDefinition.setColumnType(ColumnType.STRING);
```
DwayneJengSage: ddbName and columnType not needed here. (Similarly in the next 4 columns.)

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +217,123 @@ public void shouldConvertFreeformTextToAttachment() {
...
+        final String testStringName = "test_string";
+        final String testStringSetName = "test_string_set";
+        final String testDateName = "test_date";
+        final String testExporterDateName = "test_exporter_date";
```
DwayneJengSage: Minor: Exporter date doesn't come from DDB record. We should remove this.

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +217,123 @@ public void shouldConvertFreeformTextToAttachment() {
...
+        // setup mock
+        ExportTask mockTask = mock(ExportTask.class);
+        when(mockTask.getExporterDate()).thenReturn(TEST_LOCAL_DATE);
+
+        // create mock record
+        Item testRecord = new Item();
+        testRecord.withString(testStringName, "test_string_value");
+        testRecord.withStringSet(testStringSetName, new String[]{"test_string_set_value_1", "test_string_set_value_2"});
+        testRecord.withLong(testDateName, 1484181511);
+        testRecord.withString(testExporterDateName, "test_exporter_date_value");
```
DwayneJengSage: Also remove this.

src/test/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtilTest.java
```diff
@@ -202,4 +217,123 @@ public void shouldConvertFreeformTextToAttachment() {
...
+        List<ColumnDefinition> testColumnDefinitions;
+
+        ImmutableList.Builder<ColumnDefinition> columnDefinitionBuilder = ImmutableList.builder();
+
+        ColumnDefinition testDefinition1 = new ColumnDefinition();
+        testDefinition1.setName(testStringName);
+        testDefinition1.setColumnType(ColumnType.STRING);
+        testDefinition1.setMaximumSize(36L);
+        testDefinition1.setTransferMethod(TransferMethod.STRING);
+        testDefinition1.setDdbName(testStringName);
```
DwayneJengSage, line 295: Similarly, remove columnType from here and from the other 3 columns.

src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -47,8 +46,13 @@
     private static final Logger LOG = LoggerFactory.getLogger(SynapseExportHandler.class);
 
     // Package-scoped to be available to unit tests.
```
DwayneJengSage: Minor: Remove obsolete comment.

src/main/java/org/sagebionetworks/bridge/exporter/handler/SynapseExportHandler.java
```diff
@@ -47,8 +46,13 @@
...
-    static final List<ColumnModel> COMMON_COLUMN_LIST;
-    static {
+    private List<ColumnModel> commonColumnList;
+
+    private List<ColumnDefinition> columnDefinition;
+
+    private void setSynapseColumnDefinitionsAndColumnList() {
```
DwayneJengSage: Minor: Rename this to something like "initSynapseColumnDefinitionsAndColumnList", so we don't get it confused with an actual setter.

DwayneJengSage (changes requested):

DwayneJengSage (approved):

DwayneJengSage:
Looks good. We just need to resolve conflicts before we can merge.

PR #40, opened 2017-01-11
Limit Redrives

DwayneJengSage:
Currently, BridgeEX can redrive infinitely. This is bad if for whatever reason, a particular record or table can't be redriven.

This change adds a redrive count to each request and limits the total number of redrives based on configuration.

Testing done:
* manually tested successful export and tested injecting errors and verifying redrive max count logic

See also https://sagebionetworks.jira.com/browse/BRIDGE-1635


alxdarksage:
Looks good, I can merge when you get in.

What happens once retrieve count has been met, would we try again on the next export of a time range that includes this record?

DwayneJengSage:
> What happens once retrieve count has been met, would we try again on the next export of a time range that includes this record?

Yes, but at that point, it wouldn't be a redrive, but rather an explicit custom export from the end user.

PR #48, opened 2017-03-23
instantExportingV1

zaclinsage:
unit tests done;
manual tests done:
1. daily/hourly normal cases w/wt lastExportDateTime;
2. daily/hourly normal cases w/wt ignoreLastExportDateTime;
3. re-export;
(did not test INSTANT since we didn't use it in v1)

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/config/SpringConfig.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/config/SpringConfig.java
@@ -9,19 +9,18 @@ import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
+import com.amazonaws.auth.profile.ProfileCredentialsProvider;
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient;
 import com.amazonaws.services.dynamodbv2.document.DynamoDB;
 import com.amazonaws.services.dynamodbv2.document.Index;
 import com.amazonaws.services.dynamodbv2.document.Table;
 import com.amazonaws.services.s3.AmazonS3Client;
 import com.amazonaws.services.sqs.AmazonSQSClient;
-
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
-import org.sagebionetworks.bridge.exporter.synapse.ColumnDefinition;
-import org.sagebionetworks.bridge.json.DefaultObjectMapper;
-import org.sagebionetworks.repo.model.table.ColumnModel;
+import org.sagebionetworks.client.SynapseAdminClientImpl;
+import org.sagebionetworks.client.SynapseClient;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
@@ -32,17 +31,17 @@ import org.sagebionetworks.bridge.config.PropertiesConfig;
 import org.sagebionetworks.bridge.dynamodb.DynamoQueryHelper;
 import org.sagebionetworks.bridge.exporter.notification.S3EventNotificationCallback;
 import org.sagebionetworks.bridge.exporter.request.BridgeExporterSqsCallback;
+import org.sagebionetworks.bridge.exporter.synapse.ColumnDefinition;
 import org.sagebionetworks.bridge.exporter.util.BridgeExporterUtil;
 import org.sagebionetworks.bridge.file.FileHelper;
 import org.sagebionetworks.bridge.heartbeat.HeartbeatLogger;
+import org.sagebionetworks.bridge.json.DefaultObjectMapper;
 import org.sagebionetworks.bridge.rest.ClientManager;
 import org.sagebionetworks.bridge.rest.model.ClientInfo;
 import org.sagebionetworks.bridge.rest.model.SignIn;
 import org.sagebionetworks.bridge.s3.S3Helper;
 import org.sagebionetworks.bridge.sqs.PollSqsWorker;
 import org.sagebionetworks.bridge.sqs.SqsHelper;
-import org.sagebionetworks.client.SynapseAdminClientImpl;
-import org.sagebionetworks.client.SynapseClient;
 
 // These configs get credentials from the default credential chain. For developer desktops, this is ~/.aws/credentials.
 // For EC2 instances, this happens transparently.
@@ -136,6 +135,11 @@ public class SpringConfig {
         return ddbClient().getTable(ddbPrefix() + "Study");
     }
 
+    @Bean(name = "ddbExportTimeTable")
+    public Table ddbExportTimeTable() {
+        return ddbClient().getTable(ddbPrefix() + "ExportTime");
+    }
+
     @Bean
     public FileHelper fileHelper() {
         return new FileHelper();
@@ -213,4 +217,10 @@ public class SpringConfig {
 
         return ImmutableList.copyOf(value);
     }
+
+    @Bean(name = "ddbClientScan")
+    public AmazonDynamoDBClient ddbClientScan() {
+        return new AmazonDynamoDBClient(new ProfileCredentialsProvider());
+    }
+
 }
--- a/src/main/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessor.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessor.java
@@ -2,13 +2,14 @@ package org.sagebionetworks.bridge.exporter.record;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
-
 import javax.annotation.Resource;
 
 import com.amazonaws.services.dynamodbv2.document.Item;
 import com.amazonaws.services.dynamodbv2.document.Table;
 import com.google.common.base.Stopwatch;
+import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 import org.sagebionetworks.client.exceptions.SynapseException;
@@ -43,6 +44,8 @@ public class BridgeExporterRecordProcessor {
     // package-scoped to be available to unit tests
     static final String CONFIG_KEY_RECORD_LOOP_DELAY_MILLIS = "record.loop.delay.millis";
     static final String CONFIG_KEY_RECORD_LOOP_PROGRESS_REPORT_PERIOD = "record.loop.progress.report.period";
+    static final String STUDY_ID = "studyId";
+    static final String LAST_EXPORT_DATE_TIME = "lastExportDateTime";
 
     // config attributes
     private int delayMillis;
@@ -51,6 +54,7 @@ public class BridgeExporterRecordProcessor {
 
     // Spring helpers
     private Table ddbRecordTable;
+    private Table ddbExportTimeTable;
     private FileHelper fileHelper;
     private MetricsHelper metricsHelper;
     private RecordFilterHelper recordFilterHelper;
@@ -70,6 +74,12 @@ public class BridgeExporterRecordProcessor {
     @Resource(name = "ddbRecordTable")
     public final void setDdbRecordTable(Table ddbRecordTable) {
         this.ddbRecordTable = ddbRecordTable;
+    }
+
+    /** DDB Export Time Table. */
+    @Resource(name = "ddbExportTimeTable")
+    final void setDdbExportTimeTable(Table ddbExportTimeTable) {
+        this.ddbExportTimeTable = ddbExportTimeTable;
     }
 
     /** File helper, used for creating and cleaning up the temp dir used to store the request's temporary files. */
@@ -147,7 +157,11 @@ public class BridgeExporterRecordProcessor {
 
         Stopwatch stopwatch = Stopwatch.createStarted();
         try {
-            Iterable<String> recordIdIterable = recordIdSourceFactory.getRecordSourceForRequest(request);
+            RecordAndStudyListHolder recordAndStudyListHolder = recordIdSourceFactory.getRecordSourceForRequest(request);
+            Iterable<String> recordIdIterable = recordAndStudyListHolder.getRecordIdSource();
+            Map<String, DateTime> studyIdsToUpdate = recordAndStudyListHolder.getStudyIdsToQuery();
+            long endDateTimeEpoch = recordAndStudyListHolder.getEndDateTime().getMillis();
+
             for (String oneRecordId : recordIdIterable) {
                 // Count total number of records. Also, log at regular intervals, so people tailing the logs can follow
                 // progress.
@@ -193,6 +207,13 @@ public class BridgeExporterRecordProcessor {
 
             // We made it to the end. Set the success flag on the task.
             setTaskSuccess(task);
+
+            // update ddb exportTimeTable
+            if (!studyIdsToUpdate.isEmpty()) {
+                for (String studyId: studyIdsToUpdate.keySet()) {
+                    ddbExportTimeTable.putItem(new Item().withPrimaryKey(STUDY_ID, studyId).withNumber(LAST_EXPORT_DATE_TIME, endDateTimeEpoch));
+                }
+            }
         } finally {
             long elapsedTime = stopwatch.elapsed(TimeUnit.SECONDS);
             if (task.isSuccess()) {
--- a/src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
@@ -1,0 +1,28 @@
+package org.sagebionetworks.bridge.exporter.record;
+
+import org.joda.time.DateTime;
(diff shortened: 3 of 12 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/config/SpringConfig.java
```diff
@@ -213,4 +217,10 @@ public ExecutorService workerExecutorService() {
 
         return ImmutableList.copyOf(value);
     }
+
+    @Bean(name = "ddbClientScan")
+    public AmazonDynamoDBClient ddbClientScan() {
+        return new AmazonDynamoDBClient(new ProfileCredentialsProvider());
+    }
```
DwayneJengSage, line 224: Why do we need a DDB client? Why are we using the ProfileCredentialsProvider instead of the default one?

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactory.java
```diff
@@ -83,7 +113,7 @@ final void setS3Helper(S3Helper s3Helper) {
      * @throws IOException
      *         if we fail reading the underlying source
      */
-    public Iterable<String> getRecordSourceForRequest(BridgeExporterRequest request) throws IOException {
+    public RecordAndStudyListHolder getRecordSourceForRequest(BridgeExporterRequest request) throws IOException {
```
DwayneJengSage, line 116: Why are we changing the return type? Reading the logic, it looks like getting the iterator for record IDs is entirely independent of getting the list of study IDs. Why are we combining the logic into a single class? This adds a bunch of extra complexity for no gain.

src/main/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessor.java
```diff
@@ -193,6 +207,13 @@ public void processRecordsForRequest(BridgeExporterRequest request) throws IOExc
 
             // We made it to the end. Set the success flag on the task.
             setTaskSuccess(task);
+
+            // update ddb exportTimeTable
+            if (!studyIdsToUpdate.isEmpty()) {
+                for (String studyId: studyIdsToUpdate.keySet()) {
+                    ddbExportTimeTable.putItem(new Item().withPrimaryKey(STUDY_ID, studyId).withNumber(LAST_EXPORT_DATE_TIME, endDateTimeEpoch));
+                }
+            }
```
DwayneJengSage, line 216: We should move this into its own helper class, so that RecordProcessor isn't a big monolithic class that does everything.

Also, what about error handling? What happens when one of these puts fail?

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordAndStudyListHolder.java
```diff
@@ -0,0 +1,41 @@
...
+        this.recordIdSource = recordIdSource;
+    }
+
+    public Map<String, DateTime> getStudyIdsToQuery() {
+        return studyIdsToQuery;
+    }
+
+    public void setStudyIdsToQuery(Map<String, DateTime> studyIdsToQuery) {
+        this.studyIdsToQuery = studyIdsToQuery;
+    }
```
DwayneJengSage, line 32: Why is this a map and not a set or a list? It looks like the DateTime is not being used anywhere.

src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
```diff
@@ -0,0 +1,28 @@
...
+            return endDateTime.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0);
+        }
+    };
+
+    /**
+     * Helper method to get the startDateTime given endDateTime determining by the type of the request
+     * @param endDateTime
+     * @return
+     */
+    public abstract DateTime getStartDateTime(DateTime endDateTime);
```
DwayneJengSage, line 27: It looks like this isn't being used anywhere.

src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
```diff
@@ -0,0 +1,28 @@
...
+    },
+    HOURLY {
+        public DateTime getStartDateTime(DateTime endDateTime) {
+            return endDateTime.minusHours(1).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0);
+        }
+    },
+    INSTANT {
+        public DateTime getStartDateTime(DateTime endDateTime) {
+            return endDateTime.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0);
+        }
```
DwayneJengSage, line 19: What does this mean? Can we even calculate startDateTime from endDateTime for INSTANT? If we need to consult the lastExportTime table, should this logic live in the RecordIdSourceFactory instead?

src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
```diff
@@ -0,0 +1,28 @@
+package org.sagebionetworks.bridge.exporter.record;
+
+import org.joda.time.DateTime;
+
+public enum ExportType {
+    DAILY {
+        public DateTime getStartDateTime(DateTime endDateTime) {
+            return endDateTime.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0);
+        }
```
DwayneJengSage, line 9: This is only true if endDateTime has time 23:59:59.999. Can we enforce that that's always the case?

Also, you should use DateTime.withTimeAtStartOfDay() instead.

src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
```diff
@@ -0,0 +1,28 @@
...
+    },
+    HOURLY {
+        public DateTime getStartDateTime(DateTime endDateTime) {
+            return endDateTime.minusHours(1).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0);
+        }
```
DwayneJengSage, line 14: What happens if I submit an endTime such as 2:45pm? This code will give me a startTime of 1:00pm. Is that expected?

What happens if I submit an endTime of 2:00pm and get a startTime of 1:00pm? What happens if I then submit an endTime of 3:00pm and get a startTime of 2:00pm? There's going to be an overlap at exactly 2:00:00.000. If there are any uploads that are timestamped with that exact millisecond, they will be exported twice.

src/test/java/org/sagebionetworks/bridge/exporter/record/ExportTypeTest.java
```diff
@@ -0,0 +1,27 @@
...
+        assertEquals(startDateTime, END_DATE_TIME.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0));
+
+        testExportType = ExportType.HOURLY;
+        startDateTime = testExportType.getStartDateTime(END_DATE_TIME);
+        assertEquals(startDateTime, END_DATE_TIME.minusHours(1).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0));
+
+        testExportType = ExportType.INSTANT;
+        startDateTime = testExportType.getStartDateTime(END_DATE_TIME);
+        assertEquals(startDateTime, END_DATE_TIME.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0));
+    }
```
DwayneJengSage, line 25: This is a facetious test. Your code and your test run the exact same code. If there's a bug in the code, then there's a bug in the test.

Instead, you should test against the actual expected output. Construct a DateTime with the result you expect and test against that.

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -27,6 +42,24 @@
...
+    private static final String UPLOAD_START_DATE_TIME = "2016-05-09T00:00:00.000-0700";
+    private static final String UPLOAD_END_DATE_TIME = "2016-05-09T23:59:59.999-0700";
+    private static final String EXPORT_TIME_TABLE_NAME = "exportTime";
+    private static final String STUDY_TABLE_NAME = "Study";
+    private static final String CURRENT_DATE = "2017-03-22";
+
+    private static final DateTime UPLOAD_START_DATE_TIME_OBJ = DateTime.parse(UPLOAD_START_DATE_TIME);
+    private static final DateTime UPLOAD_END_DATE_TIME_OBJ = DateTime.parse(UPLOAD_END_DATE_TIME);
+    private static final LocalDate UPLOAD_DATE_OBJ = LocalDate.parse(UPLOAD_DATE);
+    private static final DateTime CURRENT_DATE_OBJ = DateTime.parse(CURRENT_DATE);
```
DwayneJengSage, line 55: Current date seems a little ambiguous. You start with a calendar date, but parse it as a DateTime.

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -324,26 +360,23 @@ public BridgeExporterRequest build() {
...
-            int numRecordSources = 0;
-            if (hasStartDateTime) {
-                numRecordSources++;
-            }
-            if (date != null) {
-                numRecordSources++;
+            int numSources = 0;
+            if (hasEndDateTime || date != null) {
+                numSources++;
             }
```
DwayneJengSage, line 372: Allowing both date and start/endDateTime is ambiguous.

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -161,4 +229,177 @@ private static Config mockConfig() {
...
+    public void fromDdbWithWhitelistIgnoreLastExportTime() throws Exception {
+        fromDdb(true, true, true, ExportType.DAILY);
+        fromDdb(true, false, true, ExportType.DAILY);
+        fromDdb(true, true, true, ExportType.HOURLY);
+        fromDdb(true, false, true, ExportType.HOURLY);
+        fromDdb(true, true, true, ExportType.INSTANT);
+        fromDdb(true, false, true, ExportType.INSTANT);
+    }
+
+    private static void fromDdb(boolean ignoreLastExportTime, boolean withLastExportDateTime, boolean withWhitelist, ExportType exportType) throws Exception{
```
DwayneJengSage, line 282: Can we refactor the logic so that ignoreLastExportTime, lastExportDateTime, whiteList, and exportType are independently handled? Combining them together into one big monolithic class means we have to test all 24 combinations, which is an untenable solution.

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -161,4 +229,177 @@ private static Config mockConfig() {
...
+                    assertEquals(startDateTime, CURRENT_DATE_OBJ.minusDays(1));
+                }
+            }
+        }
+
+        if (exportType == ExportType.DAILY || exportType == ExportType.HOURLY) {
+            assertEquals(retEndDateTime, UPLOAD_END_DATE_TIME_OBJ);
+        } else if (exportType == ExportType.INSTANT) {
+            assertEquals(retEndDateTime, CURRENT_DATE_OBJ.minusMinutes(1));
+        }
```
DwayneJengSage, line 403: This is also a facetious test. Basically, the code and the test are doing the same thing. If there's a bug in the code, there's a bug in the test.

What you should do instead is pass in the expected endDateTime and compare that with the result.

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactory.java
```diff
@@ -134,22 +164,74 @@ final void setS3Helper(S3Helper s3Helper) {
...
+    private Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request) {
+        DateTime endDateTime = getEndDateTime(request);
+        List<String> studyIdList = new ArrayList<>();
+
+        if (request.getStudyWhitelist() == null) {
+            // get the study id list from ddb table
+            ScanRequest scanRequest = new ScanRequest()
+                    .withTableName(ddbStudyTable.getTableName());
+
+            ScanResult result = ddbClientScan.scan(scanRequest);
```
DwayneJengSage, line 211: Why are we using the low-level API? Why can't we use the Study table directly?

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactory.java
```diff
@@ -134,22 +164,74 @@ final void setS3Helper(S3Helper s3Helper) {
...
+
+            for (Map<String, AttributeValue> item : result.getItems()) {
+                studyIdList.add(item.get(IDENTIFIER).getS());
+            }
+        } else {
+            studyIdList.addAll(request.getStudyWhitelist());
+        }
+
+        // maintain insert order for testing
+        Map<String, DateTime> studyIdsToQuery = new LinkedHashMap<>();
```
DwayneJengSage, line 221: There's nothing in the test where we validate the ordering. Also, I'm not sure we need the ordering in order to validate.

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -161,4 +229,177 @@ private static Config mockConfig() {
...
+                    .withDate(UPLOAD_DATE_OBJ)
+                    .withStudyWhitelist(studyWhitelist)
+                    .withStartDateTime(UPLOAD_START_DATE_TIME_OBJ)
+                    .withEndDateTime(UPLOAD_END_DATE_TIME_OBJ)
+                    .withIgnoreLastExportTime(ignoreLastExportTime)
+                    .withExportType(exportType)
+                    .build();
+        }
+        RecordAndStudyListHolder holder = factory.getRecordSourceForRequest(request);
+        Map<String, DateTime> studyIdsToQuery = holder.getStudyIdsToQuery();
```
DwayneJengSage, line 368: We don't test what the actual studies returned are. Also, there's no test to differentiate between getting studies from the whitelist vs getting studies from DDB. How do we know the codepaths are being tested?

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -161,4 +229,177 @@ private static Config mockConfig() {
...
+                    .withDate(UPLOAD_DATE_OBJ)
+                    .withStartDateTime(UPLOAD_START_DATE_TIME_OBJ)
+                    .withEndDateTime(UPLOAD_END_DATE_TIME_OBJ)
+                    .withIgnoreLastExportTime(ignoreLastExportTime)
+                    .withExportType(exportType)
+                    .build();
+        } else {
+            Set<String> studyWhitelist = new LinkedHashSet<>();
+            studyWhitelist.add("ddb-foo");
+            studyWhitelist.add("ddb-bar");
```
DwayneJengSage, line 357: Why does order matter here?

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -34,12 +35,14 @@
     private final Map<String, String> synapseProjectOverrideMap;
     private final Set<UploadSchemaKey> tableWhitelist;
     private final String tag;
+    private final boolean ignoreLastExportTime;
+    private final ExportType exportType;
```
DwayneJengSage, line 39: You need to add these fields to unit tests.

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -324,26 +360,23 @@ public BridgeExporterRequest build() {
                 throw new IllegalStateException("startDateTime must be before endDateTime.");
             }
 
-            // If start/endDateTime are specified, there must be a studyWhitelist.
-            if (hasStartDateTime && studyWhitelist == null) {
-                throw new IllegalStateException("If start- and endDateTime are specified, studyWhitelist must also " +
-                        "be specified.");
```
DwayneJengSage, line 330: As written, if you have a start/endDateTime, we currently still require studyWhitelist.

src/main/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManager.java
```diff
@@ -625,6 +624,7 @@ public void endOfStream(ExportTask task) throws RestartBridgeExporterException {
             }
             BridgeExporterRequest redriveRequest = new BridgeExporterRequest.Builder().copyOf(request)
                     .withTableWhitelist(redriveTableWhitelist).withRedriveCount(redriveCount + 1).withTag(redriveTag)
+                    .withIgnoreLastExportTime(true) // only redrive table needs to ignore lastExportDateTime
```
DwayneJengSage, line 627: Need to update unit tests.

PR #51, opened 2017-03-28
fix a bug relating to endDateTime and lastExportTime

zaclinsage:
also add rate limit for exportTime table
unit tests done;
manual tests done:
1. normal case for daily/hourly export;
2. special case for daily -> hourly -> daily case;
3. special case for hourly -> daily -> hourly case;

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -158,8 +158,7 @@ public class DynamoHelper {
      * @param request
      * @return
      */
-    public List<String> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime)
-            throws InterruptedException {
+    public List<String> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
         List<String> studyIdList = new ArrayList<>();
 
         if (request.getStudyWhitelist() == null) {
@@ -185,7 +184,12 @@ public class DynamoHelper {
                 studyIdsToQuery.add(studyId);
             }
 
-            TimeUnit.SECONDS.sleep(1);
+            try {
+                TimeUnit.SECONDS.sleep(1);
+            } catch (InterruptedException e) {
+                LOG.error("Unable to sleep thread: " +
+                        e.getMessage(), e);
+            }
         }
 
         return studyIdsToQuery;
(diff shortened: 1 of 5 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -162,15 +172,31 @@ public StudyInfo getStudyInfo(String studyId) {
...
+            if (studyIdItem != null) {
+                DateTime lastExportDateTime = new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone);
+                if (!endDateTime.isBefore(lastExportDateTime)) {
+                    studyIdsToQuery.add(studyId);
+                }
+            } else {
+                studyIdsToQuery.add(studyId);
+            }
+
+            TimeUnit.SECONDS.sleep(1);
```
DwayneJengSage: It would be better to swallow the InterruptedException and log a message. Otherwise, one failure would fail the entire request and force a redrive.

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -179,6 +205,8 @@ public void updateExportTimeTable(List<String> studyIdsToUpdate, DateTime endDat
                     LOG.error("Unable to update export time table for study id: " + studyId +
                             ex.getMessage(), ex);
                 }
+
+                TimeUnit.SECONDS.sleep(1);
```
DwayneJengSage: Similarly here.

src/test/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessorTest.java
```diff
@@ -177,7 +177,7 @@ public void test() throws Exception {
         // validate that we cleaned up all our files
         assertTrue(mockFileHelper.isEmpty());
 
-        verify(mockDynamoHelper).bootstrapStudyIdsToQuery(eq(REQUEST));
+        verify(mockDynamoHelper).bootstrapStudyIdsToQuery(eq(REQUEST), any());
```
DwayneJengSage: Might be better to check for the actual endDateTime, since we know what it is in the test.

DwayneJengSage (changes requested):
Mostly good. A few small comments about error handling and one test comment.

PR #52, opened 2017-04-11
V2 instant export

zaclinsage:
all unit tests done;
manual tests done;
integ tests done;

See Also:
* https://github.com/Sage-Bionetworks/BridgeJavaSDK/pull/393
* https://github.com/Sage-Bionetworks/BridgeExporterIntegrationTests/pull/1
* https://github.com/Sage-Bionetworks/BridgePF/pull/1393
* https://github.com/Sage-Bionetworks/BridgeDocs/pull/138
* https://github.com/Sage-Bionetworks/Bridge-EX-Scheduler/pull/6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>Bridge-Exporter</artifactId>
-    <version>2.0</version>
+    <version>2.1</version>
     <packaging>war</packaging>
 
     <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>Bridge-Exporter</artifactId>
-    <version>2.1</version>
+    <version>2.0</version>
     <packaging>war</packaging>
 
     <properties>
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -182,7 +182,8 @@ public class DynamoHelper {
      * @param request
      * @return
      */
-    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
+    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime)
+            throws InterruptedException {
         List<String> studyIdList = new ArrayList<>();
 
         if (request.getStudyWhitelist() == null) {
@@ -210,6 +211,9 @@ public class DynamoHelper {
                 ExportType exportType = request.getExportType();
                 studyIdsToQuery.put(studyId, exportType.getStartDateTime(endDateTime));
             }
+
+            // then sleep 1 sec before next read
+            TimeUnit.SECONDS.sleep(1);
         }
 
         return studyIdsToQuery;
@@ -220,7 +224,7 @@ public class DynamoHelper {
      * @param studyIdsToUpdate
      * @param endDateTime
      */
-    public void updateExportTimeTable(List<String> studyIdsToUpdate, DateTime endDateTime) {
+    public void updateExportTimeTable(List<String> studyIdsToUpdate, DateTime endDateTime) throws InterruptedException {
         if (!studyIdsToUpdate.isEmpty() && endDateTime != null) {
             for (String studyId: studyIdsToUpdate) {
                 try {
@@ -229,6 +233,9 @@ public class DynamoHelper {
                     LOG.error("Unable to update export time table for study id: " + studyId +
                             ex.getMessage(), ex);
                 }
+
+                // sleep 1 sec before next write
+                TimeUnit.SECONDS.sleep(1);
             }
         }
     }
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -183,7 +183,7 @@ public class DynamoHelper {
      * @return
      */
     public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime)
-            throws InterruptedException {
+             {
         List<String> studyIdList = new ArrayList<>();
 
         if (request.getStudyWhitelist() == null) {
@@ -213,7 +213,12 @@ public class DynamoHelper {
             }
 
             // then sleep 1 sec before next read
-            TimeUnit.SECONDS.sleep(1);
+            try {
+                TimeUnit.SECONDS.sleep(1);
+            } catch (InterruptedException e) {
+                LOG.error("Unable to sleep thread: " +
+                        e.getMessage(), e);
+            }
         }
 
         return studyIdsToQuery;
@@ -224,7 +229,7 @@ public class DynamoHelper {
      * @param studyIdsToUpdate
      * @param endDateTime
      */
-    public void updateExportTimeTable(List<String> studyIdsToUpdate, DateTime endDateTime) throws InterruptedException {
+    public void updateExportTimeTable(List<String> studyIdsToUpdate, DateTime endDateTime) {
         if (!studyIdsToUpdate.isEmpty() && endDateTime != null) {
             for (String studyId: studyIdsToUpdate) {
                 try {
@@ -235,7 +240,12 @@ public class DynamoHelper {
                 }
 
                 // sleep 1 sec before next write
-                TimeUnit.SECONDS.sleep(1);
+                try {
+                    TimeUnit.SECONDS.sleep(1);
+                } catch (InterruptedException e) {
+                    LOG.error("Unable to sleep thread: " +
+                            e.getMessage(), e);
+                }
             }
         }
     }
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -159,48 +159,6 @@ public class DynamoHelper {
     }
 
     /**
-     * Helper method to generate study ids for query
-     * @param request
-     * @return
-     */
-    public List<String> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
-        List<String> studyIdList = new ArrayList<>();
-
-        if (request.getStudyWhitelist() == null) {
-            // get the study id list from ddb table
-            Iterable<Item> scanOutcomes = ddbScanHelper.scan(ddbStudyTable);
-            for (Item item: scanOutcomes) {
-                studyIdList.add(item.getString(IDENTIFIER));
-            }
-        } else {
-            studyIdList.addAll(request.getStudyWhitelist());
-        }
-
-        List<String> studyIdsToQuery = new ArrayList<>();
-
-        for (String studyId : studyIdList) {
-            Item studyIdItem = ddbExportTimeTable.getItem(STUDY_ID, studyId);
-            if (studyIdItem != null) {
-                DateTime lastExportDateTime = new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone);
-                if (!endDateTime.isBefore(lastExportDateTime)) {
-                    studyIdsToQuery.add(studyId);
-                }
-            } else {
-                studyIdsToQuery.add(studyId);
-            }
-
-            try {
-                TimeUnit.SECONDS.sleep(1);
-            } catch (InterruptedException e) {
-                LOG.error("Unable to sleep thread: " +
-                        e.getMessage(), e);
-            }
-        }
-
-        return studyIdsToQuery;
-    }
-
-    /**
      * Helper method to generate study ids for query.
      * @param request
      * @return A Map with key is study id, value is the start date time to query ddb table.
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -2,6 +2,7 @@ package org.sagebionetworks.bridge.exporter.dynamo;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
@@ -11,13 +12,17 @@ import com.amazonaws.services.dynamodbv2.document.Item;
 import com.amazonaws.services.dynamodbv2.document.Table;
 import com.jcabi.aspects.Cacheable;
 import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import org.sagebionetworks.bridge.config.Config;
 import org.sagebionetworks.bridge.dynamodb.DynamoScanHelper;
+import org.sagebionetworks.bridge.exporter.record.ExportType;
 import org.sagebionetworks.bridge.exporter.request.BridgeExporterRequest;
+import org.sagebionetworks.bridge.exporter.util.BridgeExporterUtil;
 import org.sagebionetworks.bridge.json.DefaultObjectMapper;
 
 /**
@@ -42,6 +47,13 @@ public class DynamoHelper {
     private Table ddbStudyTable;
     private Table ddbExportTimeTable;
     private DynamoScanHelper ddbScanHelper;
+    private DateTimeZone timeZone;
+
+    /** Config, used to get S3 bucket for record ID override files. */
+    @Autowired
+    final void setConfig(Config config) {
+        timeZone = DateTimeZone.forID(config.get(BridgeExporterUtil.CONFIG_KEY_TIME_ZONE_NAME));
+    }
 
     /** Participant options table, used to get user sharing scope. */
     @Resource(name = "ddbParticipantOptionsTable")
@@ -163,6 +175,41 @@ public class DynamoHelper {
         }
 
         return studyIdList;
+    }
+
+    /**
+     * Helper method to generate study ids for query
+     * @param request
+     * @return
+     */
+    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
+        List<String> studyIdList = new ArrayList<>();
+
+        if (request.getStudyWhitelist() == null) {
+            // get the study id list from ddb table
+            Iterable<Item> scanOutcomes = ddbScanHelper.scan(ddbStudyTable);
+            for (Item item: scanOutcomes) {
+                studyIdList.add(item.getString(IDENTIFIER));
+            }
+        } else {
+            studyIdList.addAll(request.getStudyWhitelist());
+        }
+
+        // maintain insert order for testing
+        Map<String, DateTime> studyIdsToQuery = new HashMap<>();
+
+        for (String studyId : studyIdList) {
+            Item studyIdItem = ddbExportTimeTable.getItem(STUDY_ID, studyId);
+            if (studyIdItem != null && !request.getIgnoreLastExportTime()) {
+                studyIdsToQuery.put(studyId, new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone));
+            } else {
+                // bootstrap startDateTime with the exportType in request
+                ExportType exportType = request.getExportType();
+                studyIdsToQuery.put(studyId, exportType.getStartDateTime(endDateTime));
+            }
+        }
+
+        return studyIdsToQuery;
     }
 
     /**
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -201,7 +201,10 @@ public class DynamoHelper {
         for (String studyId : studyIdList) {
             Item studyIdItem = ddbExportTimeTable.getItem(STUDY_ID, studyId);
             if (studyIdItem != null && !request.getIgnoreLastExportTime()) {
-                studyIdsToQuery.put(studyId, new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone));
+                DateTime lastExportDateTime = new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone);
+                if (!endDateTime.isBefore(lastExportDateTime)) {
+                    studyIdsToQuery.put(studyId, lastExportDateTime);
+                }
             } else {
                 // bootstrap startDateTime with the exportType in request
                 ExportType exportType = request.getExportType();
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -202,7 +202,6 @@ public class DynamoHelper {
             studyIdList.addAll(request.getStudyWhitelist());
         }
 
-        // maintain insert order for testing
         Map<String, DateTime> studyIdsToQuery = new HashMap<>();
 
         for (String studyId : studyIdList) {
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -10,7 +10,9 @@ import javax.annotation.Resource;
 
 import com.amazonaws.services.dynamodbv2.document.Item;
 import com.amazonaws.services.dynamodbv2.document.Table;
+import com.google.common.collect.ImmutableMap;
 import com.jcabi.aspects.Cacheable;
+import org.apache.commons.lang.StringUtils;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.slf4j.Logger;
@@ -182,8 +184,12 @@ public class DynamoHelper {
      * @param request
      * @return
      */
-    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime)
-             {
+    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
+        // first check if it is s3 override request
+        if (StringUtils.isNotBlank(request.getRecordIdS3Override())) {
+            return ImmutableMap.of();
+        }
+
         List<String> studyIdList = new ArrayList<>();
 
         if (request.getStudyWhitelist() == null) {
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -180,9 +180,9 @@ public class DynamoHelper {
     }
 
     /**
-     * Helper method to generate study ids for query
+     * Helper method to generate study ids for query.
      * @param request
-     * @return
+     * @return A Map with key is study id, value is the start date time to query ddb table.
      */
     public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
         // first check if it is s3 override request
--- a/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
@@ -212,9 +212,14 @@ public class DynamoHelper {
                     studyIdsToQuery.put(studyId, lastExportDateTime);
                 }
             } else {
-                // bootstrap startDateTime with the exportType in request
-                ExportType exportType = request.getExportType();
-                studyIdsToQuery.put(studyId, exportType.getStartDateTime(endDateTime));
+                if (request.getStartDateTime() != null) {
+                    // if we setup a start date time, just use it -- normally we will not setup this field.
+                    studyIdsToQuery.put(studyId, request.getStartDateTime());
+                } else {
+                    // bootstrap startDateTime with the exportType in request
+                    ExportType exportType = request.getExportType();
+                    studyIdsToQuery.put(studyId, exportType.getStartDateTime(endDateTime));
+                }
             }
 
             // then sleep 1 sec before next read
--- a/src/main/java/org/sagebionetworks/bridge/exporter/helper/ExportHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/helper/ExportHelper.java
@@ -21,6 +21,7 @@ import org.springframework.stereotype.Component;
 
 import org.sagebionetworks.bridge.config.Config;
 import org.sagebionetworks.bridge.exporter.exceptions.BridgeExporterException;
+import org.sagebionetworks.bridge.exporter.record.ExportType;
 import org.sagebionetworks.bridge.exporter.request.BridgeExporterRequest;
 import org.sagebionetworks.bridge.exporter.synapse.SynapseHelper;
 import org.sagebionetworks.bridge.exporter.util.BridgeExporterUtil;
@@ -228,17 +229,13 @@ public class ExportHelper {
      * @return
      */
     public DateTime getEndDateTime(BridgeExporterRequest request) {
-        DateTime endDateTime;
-        if (request.getDate() != null) {
-            endDateTime = request.getDate().toDateTimeAtStartOfDay(timeZone)
-                    .withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59).withMillisOfSecond(999);
-        } else if (request.getEndDateTime() != null) {
-            endDateTime = request.getEndDateTime();
-        } else {
-            // s3 override case
-            endDateTime = null;
-        }
+        ExportType exportType = request.getExportType();
 
-        return endDateTime;
+        if (exportType == ExportType.INSTANT) {
+            // set end date time to 1 min ago to avoid clock skew issue for instant export
+            return DateTime.now().minusMinutes(1);
+        } else {
+            return request.getEndDateTime();
+        }
     }
 }
--- a/src/main/java/org/sagebionetworks/bridge/exporter/helper/ExportHelper.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/helper/ExportHelper.java
@@ -231,7 +231,7 @@ public class ExportHelper {
     public DateTime getEndDateTime(BridgeExporterRequest request) {
         ExportType exportType = request.getExportType();
 
-        if (exportType == ExportType.INSTANT) {
+        if (exportType != null && exportType == ExportType.INSTANT) {
             // set end date time to 1 min ago to avoid clock skew issue for instant export
             return DateTime.now().minusMinutes(1);
         } else {
--- a/src/main/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessor.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessor.java
@@ -164,6 +164,7 @@ public class BridgeExporterRecordProcessor {
 
         Stopwatch stopwatch = Stopwatch.createStarted();
         try {
+            // get end date time used in getting study ids and their corresponding start date time
             DateTime endDateTime = exportHelper.getEndDateTime(request);
             Map<String, DateTime> studyIdsToQuery = dynamoHelper.bootstrapStudyIdsToQuery(request, endDateTime);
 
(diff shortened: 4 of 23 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
 
     <groupId>org.sagebionetworks</groupId>
     <artifactId>Bridge-Exporter</artifactId>
-    <version>2.0</version>
+    <version>2.1</version>
```
DwayneJengSage: As stated in the other code review, bumping the version number is superfluous, since no one should be consuming the Bridge Exporter as a dependency.

Plus, we don't actually publish this to any repository, so there would be nothing to consume anyway.

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -196,6 +201,57 @@ public StudyInfo getStudyInfo(String studyId) {
     }
 
     /**
+     * Helper method to generate study ids for query
+     * @param request
+     * @return
+     */
+    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
```
DwayneJengSage: Why did we create a new one instead of modifying the existing one? Also, they have the same args. How does this even compile?

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -196,6 +201,57 @@ public StudyInfo getStudyInfo(String studyId) {
     }
 
     /**
+     * Helper method to generate study ids for query
+     * @param request
+     * @return
+     */
```
DwayneJengSage: Needs better docs. What does this return? What does the DateTime in the map represent?

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -196,6 +201,57 @@ public StudyInfo getStudyInfo(String studyId) {
...
+    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
+        // first check if it is s3 override request
+        if (StringUtils.isNotBlank(request.getRecordIdS3Override())) {
+            return ImmutableMap.of();
+        }
```
DwayneJengSage, line 170: Need a unit test for this case.

src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
```diff
@@ -0,0 +1,28 @@
+package org.sagebionetworks.bridge.exporter.record;
+
+import org.joda.time.DateTime;
+
+public enum ExportType {
+    DAILY {
+        public DateTime getStartDateTime(DateTime endDateTime) {
+            return endDateTime.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0);
+        }
```
DwayneJengSage: There's some wonkiness with time. You'll want to use DateTime.withTimeAtStartOfDay() instead. (Here and INSTANT.)

src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
```diff
@@ -338,25 +358,101 @@ public void bootstrapStudyIdsToQueryTestNoItemInExportTimeTable() {
...
+        request = new BridgeExporterRequest.Builder().withEndDateTime(UPLOAD_END_DATE_TIME_OBJ)
+                .withExportType(ExportType.HOURLY)
+                .withStudyWhitelist(ImmutableSet.of("ddb-foo"))
+                .build();
+
+        studyIdsToUpdate = dynamoHelper.bootstrapStudyIdsToQuery(request, UPLOAD_END_DATE_TIME_OBJ);
+
+        assertEquals(studyIdsToUpdate.size(), 1);
+        assertEquals(studyIdsToUpdate.get("ddb-foo").getMillis(), UPLOAD_END_DATE_TIME_OBJ
+                .minusHours(1).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0).getMillis());
```
DwayneJengSage: This test doesn't actually test that the result is correct, just that it runs the same code. If the code is wrong, the test is wrong too.

Better to actually test against the result we're expecting.

(Also in bootstrapStudyIdsToQueryTestIgnoreLastExportTime().)

src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
```diff
@@ -295,29 +299,44 @@ public void bootstrapStudyIdsToQueryTestNormal() {
...
-        request = new BridgeExporterRequest.Builder().withDate(UPLOAD_DATE_OBJ).build();
+        request = new BridgeExporterRequest.Builder().withEndDateTime(UPLOAD_END_DATE_TIME_OBJ)
+                .withExportType(ExportType.DAILY).build();
         studyIdsToUpdate = dynamoHelper.bootstrapStudyIdsToQuery(request, UPLOAD_END_DATE_TIME_OBJ);
 
         assertEquals(studyIdsToUpdate.size(), 2);
-        assertEquals(studyIdsToUpdate.get(0), "ddb-foo");
-        assertEquals(studyIdsToUpdate.get(1), "ddb-bar");
+        assertEquals(studyIdsToUpdate.get("ddb-foo").getMillis(), UPLOAD_START_DATE_TIME_OBJ.getMillis());
+        assertEquals(studyIdsToUpdate.get("ddb-bar").getMillis(), UPLOAD_START_DATE_TIME_OBJ.getMillis());
```
DwayneJengSage: It would be better if the lastExportTime wasn't exactly the same as daily. Otherwise, we have no way of knowing whether we got this result because of lastExportTime or if we got this result because we ignored lastExportTime and calculated the start date time.

(Similarly for hourly and instant, and similarly for the null item and ignoreLastExportTime cases.)

src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
```diff
@@ -338,25 +358,101 @@ public void bootstrapStudyIdsToQueryTestNoItemInExportTimeTable() {
...
+
+    @Test
+    public void bootstrapStudyIdsToQueryTestIgnoreLastExportTime() {
+
+        // mock study table and study id list
+        Table mockStudyTable = mock(Table.class);
+        DynamoScanHelper mockDdbScanHelper = mock(DynamoScanHelper.class);
+
+        Item item1 = new Item().withString(IDENTIFIER, "ddb-foo");
+        Item item2 = new Item().withString(IDENTIFIER, "ddb-bar");
```
DwayneJengSage: Our DDB items should include lastExportTime. Otherwise, we have no way of knowing whether we got the result because we saw lastExportTime and ignored it or whether we didn't see lastExportTime at all.

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -311,39 +283,43 @@ public Builder withTag(String tag) {
...
+                    throw new IllegalStateException("Must specify endDateTime for daily or hourly export.");
+                }
+
+                if (exportType == ExportType.INSTANT && hasEndDateTime) {
+                    throw new IllegalStateException("Must not specify endDateTime for instant export.");
+                }
+
+                if (exportType == ExportType.INSTANT && studyWhitelist == null) {
+                    throw new IllegalStateException("Must specify study whitelist for instant export.");
+                }
```
DwayneJengSage, line 361: Also, hourly.

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -107,16 +95,6 @@ public BridgeExporterSharingMode getSharingMode() {
...
 
     /**
-     * Start date, inclusive. For use with export jobs more granular than daily (such as hourly exports). See also
-     * {@link #getEndDateTime}. If this is specified, so must startDateTime. You must specify exactly one of date,
-     * start/endDateTime, or recordIdS3Override.
-     */
-    @JsonSerialize(using = DateTimeToStringSerializer.class)
-    public DateTime getStartDateTime() {
-        return startDateTime;
-    }
```
DwayneJengSage, line 117: I think we should keep startDateTime. This is useful for testing, for redrives, and for re-exports. (Obviously can only be called with ignoreLastExportTime.)

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -149,6 +127,14 @@ public String getTag() {
         return tag;
     }
 
+    public boolean getIgnoreLastExportTime() {
+        return this.ignoreLastExportTime;
+    }
```
DwayneJengSage, line 143: We should probably enforce that ignoreLastExportTime must be used with exporterDdbPrefixOverride, recordIdS3Override, synapseProjectOverrideMap, or tableWhitelist, since these are always non-standard export jobs and should neither use lastExportTime nor update lastExportTime.

src/test/java/org/sagebionetworks/bridge/exporter/helper/ExportHelperTest.java
```diff
@@ -87,20 +101,31 @@ public void getEndDateTimeTest() {
         exportHelper.setConfig(mockConfig);
 
         // DAILY
-        request = new BridgeExporterRequest.Builder().withDate(UPLOAD_DATE_OBJ).build();
+        request = new BridgeExporterRequest.Builder().withEndDateTime(UPLOAD_END_DATE_TIME_OBJ).withExportType(
+                ExportType.DAILY).build();
         endDateTime = exportHelper.getEndDateTime(request);
 
         assertEquals(endDateTime.getMillis(), UPLOAD_END_DATE_TIME_OBJ.getMillis());
```
DwayneJengSage: We should compare the dates themselves and not just the epoch milliseconds. This ensures were aren't losing time zone information. (Similarly for hourly and instant.)

src/test/java/org/sagebionetworks/bridge/exporter/helper/ExportHelperTest.java
```diff
@@ -87,20 +101,31 @@ public void getEndDateTimeTest() {
...
 
+        // INSTANT
+        request = new BridgeExporterRequest.Builder()
+                .withExportType(ExportType.INSTANT)
+                .withStudyWhitelist(ImmutableSet.of("dummy-whitelist"))
+                .build();
+        endDateTime = exportHelper.getEndDateTime(request);
+
+        assertEquals(endDateTime.getMillis(), UPLOAD_END_DATE_TIME_OBJ.minusMinutes(1).getMillis());
```
DwayneJengSage: This doesn't test the result. It tests that the test and the code run the same code. We need to actually check against the result we're expecting instead of trying to calculate it.

src/main/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessor.java
```diff
@@ -163,7 +164,10 @@ public void processRecordsForRequest(BridgeExporterRequest request) throws IOExc
 
         Stopwatch stopwatch = Stopwatch.createStarted();
         try {
-            Iterable<String> recordIdIterable = recordIdSourceFactory.getRecordSourceForRequest(request);
+            DateTime endDateTime = exportHelper.getEndDateTime(request);
+            Map<String, DateTime> studyIdsToQuery = dynamoHelper.bootstrapStudyIdsToQuery(request, endDateTime);
```
DwayneJengSage: More comments so we know what's going on please.

src/test/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessorTest.java
```diff
@@ -178,18 +178,41 @@ public void test() throws Exception {
...
 
-        verify(mockDynamoHelper).bootstrapStudyIdsToQuery(eq(REQUEST), eq(END_DATE_TIME));
+        verify(mockDynamoHelper).bootstrapStudyIdsToQuery(eq(REQUEST), eq(null));
+
         verify(mockDynamoHelper).updateExportTimeTable(any(), any());
         verify(mockExportHelper).getEndDateTime(eq(REQUEST));
     }
 
     @Test
+    public void testIgnoreLastExportTime() throws Exception {
```
DwayneJengSage, line 196: Need to verify that we *don't* call DynamoHelper.updateExportTimeTable()

src/test/java/org/sagebionetworks/bridge/exporter/record/ExportTypeTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class ExportTypeTest {
+    private static final String END_DATE_TIME_STR = "2016-05-09T23:59:59.999-0700";
+    private static final DateTime END_DATE_TIME = DateTime.parse(END_DATE_TIME_STR);
+
+    @Test
+    public void testGetStartDateTime() {
+        ExportType testExportType = ExportType.DAILY;
+        DateTime startDateTime = testExportType.getStartDateTime(END_DATE_TIME);
+        assertEquals(startDateTime, END_DATE_TIME.withHourOfDay(0).withMinuteOfHour(0).withSecondOfMinute(0).withMillisOfSecond(0));
```
DwayneJengSage: Similarly here, in all 3 cases, we need to test the actual result, and not just that the code and the test have the same code.

src/main/java/org/sagebionetworks/bridge/exporter/record/ExportType.java
```diff
@@ -0,0 +1,28 @@
+package org.sagebionetworks.bridge.exporter.record;
+
+import org.joda.time.DateTime;
+
+public enum ExportType {
```
DwayneJengSage, line 5: This class always assumes that DAILY endDateTime will be on the same day, but HOURLY endDateTime will be on the next hour. Which is it? If it's not consistent, this means we'll either be losing a millisecond on the DAILY case, or we have an overlapping millisecond on the HOURLY case.

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactory.java
```diff
@@ -83,64 +74,34 @@ final void setS3Helper(S3Helper s3Helper) {
...
-        // use endDateTime minus 1 millisecond.
-        long startEpochTime = startDateTime.getMillis();
-        long endEpochTime = endDateTime.getMillis() - 1;
-
-        // RangeKeyCondition can be shared across multiple queries.
-        RangeKeyCondition rangeKeyCondition = new RangeKeyCondition("uploadedOn").between(startEpochTime,
-                endEpochTime);
+    private Iterable<String> getDynamoRecordIdSourceGeneral(DateTime endDateTime, Map<String, DateTime> studyIdsToQuery) {
+        // proceed
+        Iterable<Item> recordItemIter;
```
DwayneJengSage: Minor: You don't need to declare this variable at the top when it's not used until the end.

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactory.java
```diff
@@ -83,64 +74,34 @@ final void setS3Helper(S3Helper s3Helper) {
...
         // We need to make a separate query for _each_ study in the whitelist. That's just how DDB hash keys work.
         List<Iterable<Item>> recordItemIterList = new ArrayList<>();
-        for (String oneStudyId : request.getStudyWhitelist()) {
-            Iterable<Item> recordItemIter = ddbQueryHelper.query(ddbRecordStudyUploadedOnIndex, "studyId", oneStudyId,
-                    rangeKeyCondition);
-            recordItemIterList.add(recordItemIter);
+        for (Map.Entry<String, DateTime> oneStudyIdAndDateTime : studyIdsToQuery.entrySet()) {
+            Iterable<Item> recordItemIterTemp = ddbQueryHelper.query(ddbRecordStudyUploadedOnIndex, "studyId", oneStudyIdAndDateTime.getKey(),
+                    new RangeKeyCondition("uploadedOn").between(oneStudyIdAndDateTime.getValue().getMillis(),
+                            endDateTime.getMillis()));
```
DwayneJengSage: Minor: Maybe break this statement into multiple statements so it's more readable?

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -311,39 +283,43 @@ public Builder withTag(String tag) {
             return this;
         }
 
+        public Builder withIgnoreLastExportTime(boolean ignoreLastExportTime) {
+            this.ignoreLastExportTime = ignoreLastExportTime;
+            return this;
+        }
```
DwayneJengSage, line 316: Need to add a case to unit tests, for both the builder test and the JSON test.

src/main/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManager.java
```diff
@@ -566,8 +565,8 @@ public void endOfStream(ExportTask task) throws RestartBridgeExporterException {
             } else {
                 redriveTag = REDRIVE_TAG_PREFIX + tag;
             }
-            BridgeExporterRequest redriveRequest = new BridgeExporterRequest.Builder().copyOf(request).withDate(null)
-                    .withStartDateTime(null).withEndDateTime(null).withRecordIdS3Override(filename).withTag(redriveTag)
+            BridgeExporterRequest redriveRequest = new BridgeExporterRequest.Builder().copyOf(request)
+                    .withEndDateTime(null).withExportType(null).withRecordIdS3Override(filename).withTag(redriveTag)
                     .withRedriveCount(redriveCount + 1).build();
```
DwayneJengSage: Should specify ignoreLastExportTime just to be clear and safe. Similarly for the redrive table case.

src/main/java/org/sagebionetworks/bridge/exporter/helper/ExportHelper.java
```diff
@@ -228,17 +229,13 @@ public String uploadFreeformTextAsAttachment(String recordId, String text) throw
...
-        if (request.getDate() != null) {
-            endDateTime = request.getDate().toDateTimeAtStartOfDay(timeZone)
-                    .withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59).withMillisOfSecond(999);
-        } else if (request.getEndDateTime() != null) {
-            endDateTime = request.getEndDateTime();
+        ExportType exportType = request.getExportType();
+
+        if (exportType == ExportType.INSTANT) {
+            // set end date time to 1 min ago to avoid clock skew issue for instant export
+            return DateTime.now().minusMinutes(1);
```
DwayneJengSage: Time zone?

src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
DwayneJengSage: We should also compare dates and not just epoch milliseconds, so we can make sure we're setting timezone correctly. (Here and everywhere else in this test class.)

src/test/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessorTest.java
```diff
@@ -129,7 +129,7 @@ public void test() throws Exception {
         // mock record ID factory
         List<String> recordIdList = ImmutableList.of("success-record-1", "filtered-record", "missing-record",
                 "error-record", "success-record-2");
-        when(mockRecordIdFactory.getRecordSourceForRequest(REQUEST)).thenReturn(recordIdList);
+        when(mockRecordIdFactory.getRecordSourceForRequest(same(REQUEST), any(), any())).thenReturn(recordIdList);
```
DwayneJengSage: We should validate that the correct args are being passed in.

src/test/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessorTest.java
```diff
@@ -178,18 +178,41 @@ public void test() throws Exception {
         // validate that we cleaned up all our files
         assertTrue(mockFileHelper.isEmpty());
 
-        verify(mockDynamoHelper).bootstrapStudyIdsToQuery(eq(REQUEST), eq(END_DATE_TIME));
+        verify(mockDynamoHelper).bootstrapStudyIdsToQuery(eq(REQUEST), eq(null));
```
DwayneJengSage: We need to verify the correct endDateTime is being passed in.

src/test/java/org/sagebionetworks/bridge/exporter/record/BridgeExporterRecordProcessorTest.java
```diff
@@ -178,18 +178,41 @@ public void test() throws Exception {
...
+
         verify(mockDynamoHelper).updateExportTimeTable(any(), any());
```
DwayneJengSage: We need to verify the correct inputs.

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -27,96 +27,78 @@
...
+            assertEquals(recordIdList.get(0), "foo-1");
+            assertEquals(recordIdList.get(1), "foo-2");
+            assertEquals(recordIdList.get(2), "bar-1");
+            assertEquals(recordIdList.get(3), "bar-2");
+
+            validateRangeKey(fooRangeKeyCaptor.getValue(), UPLOAD_START_DATE_TIME_OBJ.getMillis(), UPLOAD_END_DATE_TIME_OBJ.getMillis());
+            validateRangeKey(barRangeKeyCaptor.getValue(), UPLOAD_START_DATE_TIME_OBJ.getMillis(), UPLOAD_END_DATE_TIME_OBJ.getMillis());
+        } else {
+            assertEquals(recordIdList.size(), 4); // only output records in given time range
+        }
```
DwayneJengSage: I'm a little confused. How does the code vary when endTime is before lastExportTime, and how do we test for the different behaviors?

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
```diff
@@ -145,7 +127,7 @@ public void fromS3Override() throws Exception {
         // execute and validate
         BridgeExporterRequest request = new BridgeExporterRequest.Builder()
                 .withRecordIdS3Override("dummy-override-file").build();
-        Iterable<String> recordIdIter = factory.getRecordSourceForRequest(request);
+        Iterable<String> recordIdIter = factory.getRecordSourceForRequest(request, UPLOAD_END_DATE_TIME_OBJ, ImmutableMap.of());
```
DwayneJengSage: For S3 override case, wouldn't endTime be null?

src/test/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManagerEndOfStreamTest.java
```diff
@@ -269,16 +272,13 @@ public void recoverableFailures() throws Exception {
         List<BridgeExporterRequest> redriveRequestList = redriveRequestCaptor.getAllValues();
 
         BridgeExporterRequest redriveRecordRequest = redriveRequestList.get(0);
-        assertNull(redriveRecordRequest.getDate());
-        assertNull(redriveRecordRequest.getStartDateTime());
```
DwayneJengSage, line 273: Need to verify we blank out ExportType.

DwayneJengSage (changes requested):

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -196,6 +201,57 @@ public StudyInfo getStudyInfo(String studyId) {
...
+
+        // maintain insert order for testing
+        Map<String, DateTime> studyIdsToQuery = new HashMap<>();
+
+        for (String studyId : studyIdList) {
+            Item studyIdItem = ddbExportTimeTable.getItem(STUDY_ID, studyId);
+            if (studyIdItem != null && !request.getIgnoreLastExportTime()) {
+                DateTime lastExportDateTime = new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone);
+                if (!endDateTime.isBefore(lastExportDateTime)) {
```
DwayneJengSage: What happens if endDateTime is null? We just hit this error in production. We should add a null-check and test for it here.

DwayneJengSage (changes requested):

src/test/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactoryTest.java
zaclinsage: yeah you are right. and i think we should test end date time before last export date time in the method bootstrapStudyIdsToQuery() instead of this -- since we move such logic in DynamoHelper class already.

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
zaclinsage: just added

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -196,6 +201,61 @@ public StudyInfo getStudyInfo(String studyId) {
     }
 
     /**
+     * Helper method to generate study ids for query.
+     * @param request
+     * @return A Map with key is study id, value is the start date time to query ddb table.
+     */
+    public Map<String, DateTime> bootstrapStudyIdsToQuery(BridgeExporterRequest request, DateTime endDateTime) {
```
DwayneJengSage: Why did we create a new one instead of modifying the existing one? Also, they have the same args. How does this even compile?

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -7,65 +7,60 @@
...
     }
 
-    /**
-     * Date for data that Bridge EX should export (usually yesterday's date). You must specify exactly one of date,
-     * start/endDateTime, or recordIdS3Override.
-     */
-    @JsonSerialize(using = LocalDateToStringSerializer.class)
-    public LocalDate getDate() {
-        return date;
+    @JsonSerialize(using = DateTimeToStringSerializer.class)
```
DwayneJengSage: We lost the documentation for startDateTime.

src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
```diff
@@ -39,11 +41,17 @@
...
+    private static final String EXPORT_START_DATE_TIME = "2016-05-07T00:00:00.000-0700";
     private static final String UPLOAD_END_DATE_TIME = "2016-05-09T23:59:59.999-0700";
+    private static final String UPLOAD_END_DATE_TIME_HOUR = "2016-05-09T22:00:00.000-0700";
 
     private static final DateTime UPLOAD_START_DATE_TIME_OBJ = DateTime.parse(UPLOAD_START_DATE_TIME);
+    private static final DateTime EXPORT_START_DATE_TIME_OBJ = DateTime.parse(EXPORT_START_DATE_TIME);
     private static final DateTime UPLOAD_END_DATE_TIME_OBJ = DateTime.parse(UPLOAD_END_DATE_TIME);
+    private static final DateTime UPLOAD_END_DATE_TIME_OBJ_HOUR = DateTime.parse(UPLOAD_END_DATE_TIME_HOUR);
+    private static final DateTime LAST_EXPORT_DATE_TIME_OBJ = UPLOAD_START_DATE_TIME_OBJ.minus(1);
+    private static final DateTime LAST_EXPORT_DATE_TIME_OBJ_AFTER = UPLOAD_END_DATE_TIME_OBJ.plus(1);
```
DwayneJengSage: For clarity, use minusMillis() and plusMillis() instead of minus() and plus().

src/test/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelperTest.java
```diff
@@ -39,11 +41,17 @@
     private static final String EXPORT_TIME_TABLE_NAME = "exportTime";
 
     private static final String UPLOAD_DATE = "2016-05-09";
-    private static final String UPLOAD_START_DATE_TIME = "2016-05-09T00:00:00.000-0700";
+    private static final String UPLOAD_START_DATE_TIME = "2016-05-08T00:00:00.000-0700";
+    private static final String EXPORT_START_DATE_TIME = "2016-05-07T00:00:00.000-0700";
     private static final String UPLOAD_END_DATE_TIME = "2016-05-09T23:59:59.999-0700";
+    private static final String UPLOAD_END_DATE_TIME_HOUR = "2016-05-09T22:00:00.000-0700";
```
DwayneJengSage: This should be start time, not end time.

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -311,44 +305,67 @@ public Builder withTag(String tag) {
...
-            boolean hasEndDateTime = endDateTime != null;
-            if (hasStartDateTime ^ hasEndDateTime) {
-                throw new IllegalStateException("startDateTime and endDateTime must both be specified or both be " +
-                        "absent.");
-            }
-            if (hasStartDateTime && !startDateTime.isBefore(endDateTime)) {
-                throw new IllegalStateException("startDateTime must be before endDateTime.");
+            if (startDateTime != null && !ignoreLastExportTime) {
+                throw new IllegalStateException("Cannot specify start date time without setting ignore last export date time.");
             }
```
DwayneJengSage: Can probably combine this with "if (hasRecordIdS3Override || hasSynapseProjectOverrideMap ..."

src/test/java/org/sagebionetworks/bridge/exporter/helper/ExportHelperTest.java
```diff
@@ -77,32 +93,38 @@ public void uploadFreeformText() throws Exception {
...
         ExportHelper exportHelper = new ExportHelper();
-        exportHelper.setConfig(mockConfig);
 
         // DAILY
-        request = new BridgeExporterRequest.Builder().withDate(UPLOAD_DATE_OBJ).build();
+        request = new BridgeExporterRequest.Builder().withEndDateTime(UPLOAD_END_DATE_TIME_OBJ).withExportType(
+                ExportType.DAILY).build();
         endDateTime = exportHelper.getEndDateTime(request);
 
         assertEquals(endDateTime.getMillis(), UPLOAD_END_DATE_TIME_OBJ.getMillis());
```
DwayneJengSage: We should compare the dates themselves and not just the epoch milliseconds. This ensures were aren't losing time zone information. (Similarly for hourly and instant.)

src/main/java/org/sagebionetworks/bridge/exporter/helper/ExportHelper.java
DwayneJengSage: Time zone?

src/main/java/org/sagebionetworks/bridge/exporter/dynamo/DynamoHelper.java
```diff
@@ -196,6 +201,61 @@ public StudyInfo getStudyInfo(String studyId) {
...
+            studyIdList.addAll(request.getStudyWhitelist());
+        }
+
+        Map<String, DateTime> studyIdsToQuery = new HashMap<>();
+
+        for (String studyId : studyIdList) {
+            Item studyIdItem = ddbExportTimeTable.getItem(STUDY_ID, studyId);
+            if (studyIdItem != null && !request.getIgnoreLastExportTime()) {
+                DateTime lastExportDateTime = new DateTime(studyIdItem.getLong(LAST_EXPORT_DATE_TIME), timeZone);
+                if (!endDateTime.isBefore(lastExportDateTime)) {
```
DwayneJengSage: What happens if endDateTime is null? We just hit this error in production. We should add a null-check and test for it here.

src/main/java/org/sagebionetworks/bridge/exporter/record/RecordIdSourceFactory.java
```diff
@@ -83,64 +74,33 @@ final void setS3Helper(S3Helper s3Helper) {
...
+    private Iterable<String> getDynamoRecordIdSourceGeneral(DateTime endDateTime, Map<String, DateTime> studyIdsToQuery) {
         // We need to make a separate query for _each_ study in the whitelist. That's just how DDB hash keys work.
         List<Iterable<Item>> recordItemIterList = new ArrayList<>();
-        for (String oneStudyId : request.getStudyWhitelist()) {
-            Iterable<Item> recordItemIter = ddbQueryHelper.query(ddbRecordStudyUploadedOnIndex, "studyId", oneStudyId,
-                    rangeKeyCondition);
-            recordItemIterList.add(recordItemIter);
+        for (Map.Entry<String, DateTime> oneStudyIdAndDateTime : studyIdsToQuery.entrySet()) {
+            RangeKeyCondition rangeKeyCondition = new RangeKeyCondition("uploadedOn")
+                    .between(oneStudyIdAndDateTime.getValue().getMillis(), endDateTime.getMillis());
```
DwayneJengSage: This contradicts the logic in ExportType, which assumes endDateTime is always on the next hour. However, if endDateTime is on the next hour, then we always have an overlap of 1 millisecond where an upload might be uploaded twice.

Also, we always set endDateTime as lastExportTime, then use that as the next startDateTime, which means we'll have an overlap.

Finally, BridgeExporterRequest says endDateTime is exclusive, but here we use it inclusively.

src/test/java/org/sagebionetworks/bridge/exporter/worker/ExportWorkerManagerEndOfStreamTest.java
```diff
@@ -269,16 +272,14 @@ public void recoverableFailures() throws Exception {
...
+        assertNull(redriveRecordRequest.getExportType());
```
DwayneJengSage, line 275: Also need to verify that we set ignoreLastExportTime to true (here and in the redriveTableRequest).

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -311,44 +305,67 @@ public Builder withTag(String tag) {
...
+                    throw new IllegalStateException("Must specify study whitelist for hourly export.");
+                }
+
+                if (hasRecordIdS3Override) {
+                    throw new IllegalStateException("Cannot specify recordIdS3Override for daily, hourly or instant export.");
+                }
+            } else {
+                if (!hasRecordIdS3Override) {
+                    throw new IllegalStateException("Must specify recordIdS3Override for override export.");
+                }
```
DwayneJengSage: Shouldn't we allow startDateTime to not have exportType? In fact, startDateTime doesn't really make sense with exportType, since we'd neither be using lastExportTime nor would we be calculating startDateTime from exportType.

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -311,44 +305,67 @@ public Builder withTag(String tag) {
...
         public BridgeExporterRequest build() {
-            // startDateTime and endDateTime must be both specified or both absent.
-            boolean hasStartDateTime = startDateTime != null;
-            boolean hasEndDateTime = endDateTime != null;
-            if (hasStartDateTime ^ hasEndDateTime) {
-                throw new IllegalStateException("startDateTime and endDateTime must both be specified or both be " +
-                        "absent.");
-            }
-            if (hasStartDateTime && !startDateTime.isBefore(endDateTime)) {
-                throw new IllegalStateException("startDateTime must be before endDateTime.");
```
DwayneJengSage, line 324: Shouldn't we keep the check for startDateTime before endDateTime?

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -311,44 +305,67 @@ public Builder withTag(String tag) {
...
+        }
+
         /** Builds a Bridge EX request object and validates all parameters. */
         public BridgeExporterRequest build() {
-            // startDateTime and endDateTime must be both specified or both absent.
-            boolean hasStartDateTime = startDateTime != null;
-            boolean hasEndDateTime = endDateTime != null;
-            if (hasStartDateTime ^ hasEndDateTime) {
-                throw new IllegalStateException("startDateTime and endDateTime must both be specified or both be " +
-                        "absent.");
```
DwayneJengSage: Should we keep the check for startDateTime without endDateTime?

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
DwayneJengSage: What happens if you specify recordIdS3Override and endDateTime?

src/main/java/org/sagebionetworks/bridge/exporter/request/BridgeExporterRequest.java
```diff
@@ -208,44 +201,52 @@ public String toString() {
...
-        /** @see BridgeExporterRequest#getDate */
-        @JsonDeserialize(using = LocalDateDeserializer.class)
-        public Builder withDate(LocalDate date) {
-            this.date = date;
+        /** @see BridgeExporterRequest#getStartDateTime()  */
+        @JsonDeserialize(using = DateTimeDeserializer.class)
+        public Builder withStartDateTime(DateTime startDateTime) {
+            this.startDateTime = startDateTime;
             return this;
         }
```
DwayneJengSage, line 256: Need a test that successfully constructs a request with startDateTime.

DwayneJengSage (changes requested):

PR #61, opened 2017-10-19
extra rate limiting for SynapseClient.getColumnModelsForTableEntity()

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/PLFM-4673

Synapse team fixed and turned on extra throttling for expensive API calls earlier this week. This led to extra throttles in Production. I added a separate rate limiter for getColumnModels(), and an extra config so we can tweak this setting in the future without needing a code push.

This change is primarily a rate limiting change, not a functionality change, so there's not much I can do to test this specific change. However, I did run mvn verify (unit tests, FindBugs, Jacoco test coverage) and integ tests to make sure there are no regressions.


alxdarksage (commented):
Approved

PR #62, opened 2017-11-17
Deploy app to new AWS bridge-exporter environments

zaro0508:
The infra for Bridge-Exporter app is now automated with
Bridge-Exporter-infra[1] repo.  This new project uses cloudformation
to setup the bridge exporter env and SQS queues. We should switch to
using this environment and deprecate the manually created exporter
environment.

[1] https://github.com/Sage-Bionetworks/Bridge-Exporter-infra

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,11 @@ deploy:
   secret_access_key:
     secure: IVPBSeFrGy/Oe4Jgxmp5BxTCEe1IuCQTPcZcmLr5E2FzVRqq7W7sKEOT8KKvt6OlpF+MUBY6jqJVT9AYweZqw7XrqzYzzLNIwWKwEMUptLlk46gOp0H6TWHPlSBf0jSB3OpaUkIbhaXTykyflyP4pYXzWAse/FL0T4kOCMaa4Fc=
   region: us-east-1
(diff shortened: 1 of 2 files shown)


DwayneJengSage:
Can we hold off on making these changes until after I get back on Nov 27? We have monitoring for if Bridge Exporter throws a bunch of errors, but no monitoring if the Bridge Exporter silently stops working. Given that Bridge Exporter is critical both to delivering data for researchers, and also for validating data uploads for apps in development, we should probably wait until I get back so we can keep a close eye on it.

Unless @alxdarksage is comfortable enough with Bridge Exporter to do ops on it while I'm gone.

DwayneJengSage (approved):

zaro0508:
Can do.  How about you create a jira for the alarms we need to setup for Bridge-Exporter app and I'll add it to the Bridge-Exporter-infra project?

DwayneJengSage:
There are two parts. The first is replicating the existing alarms in Sumo Logic. The second is to create alarms that don't exist.

DwayneJengSage:
As per other comment, it's too risky to do this migration in this manner. We need a better migration strategy. Closing this so we don't accidentally push it out.

PR #64, opened 2017-12-24
BRIDGE-1225: monitor with new relic

zaro0508:
Setup agent to allow EC2 monitoring with new relic.
Note: only works when deployed to EB bridge-exporter-* Envs


DwayneJengSage (commented):
What happens when this is deployed to the old Exporter envs?

DwayneJengSage (approved):

PR #78, opened 2019-02-09
sanitize validation errors; fix null-pointer exception

DwayneJengSage:
We currently don't sanitize validation errors. This includes stripping away newlines, tabs, and escaping strings. This leads to really weird things happening in Synapse when we have unescaped quotes.


DwayneJengSage:
I'm going ahead and merging and deploying this, because this issue might recur in production over the weekend.

alxdarksage (commented):
Approved, though I don't really see where sanitizing the input is effected.

DwayneJengSage:
> Approved, though I don't really see where sanitizing the input is effected.

It's in ColumnDefinition.json.

PR #82, opened 2019-03-21
Use CSVWriter Library

DwayneJengSage:
https://sagebionetworks.jira.com/browse/BRIDGE-2369

Instead of manually creating our own TSVs, we should use an existing CSVWriter library. This means we're less likely to hit weird TSV formatting errors.

Testing done:
* mvn verify (unit tests, FindBugs, Jacoco test coverage)
* integ tests
* manually tested newlines, carriage returns, tabs, quotes, and escapes in every type of field and metadata

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
@@ -1,14 +1,18 @@
 package org.sagebionetworks.bridge.exporter.worker;
 
 import java.io.File;
-import java.io.PrintWriter;
+import java.io.IOException;
+import java.io.Writer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
-import com.google.common.base.Joiner;
+import au.com.bytecode.opencsv.CSVWriter;
 
 import com.google.common.collect.ImmutableList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.sagebionetworks.bridge.exporter.exceptions.BridgeExporterException;
 import org.sagebionetworks.bridge.exporter.exceptions.BridgeExporterTsvException;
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
```diff
@@ -70,11 +78,31 @@ public void checkInitAndThrow() throws BridgeExporterException {
...
+            LOG.error("Error flushing TSV writer: " + ex.getMessage(), ex);
+            throw new BridgeExporterException("Error flushing TSV writer: " + ex.getMessage(), ex);
+        } finally {
+            try {
+                tsvWriter.close();
+            } catch (IOException ex) {
+                LOG.error("Error closing TSV writer: " + ex.getMessage(), ex);
+                //noinspection ThrowFromFinallyBlock
+                throw new BridgeExporterException("Error closing TSV writer: " + ex.getMessage(), ex);
+            }
```
alxdarksage, line 104: Would try-with-resources help here? The logging of the IOException would probably not differentiate between an error on flush and an error on close, but that might be in the origin exception message.

src/test/java/org/sagebionetworks/bridge/exporter/handler/PhoneAppVersionInfoTest.java
```diff
@@ -38,12 +38,12 @@ public void normalCase() {
     @Test
     public void sanitizeFields() {
         String metadataText = "{\n" +
-                "   \"appVersion\":\"Bridge-EX\\t\\t2.0\",\n" +
-                "   \"phoneInfo\":\"Extra\\t\\ttabs\"\n" +
+                "   \"appVersion\":\"<p>Bridge-EX 2.0</p>\",\n" +
+                "   \"phoneInfo\":\"<b>formatting tag</b>\"\n" +
```
alxdarksage, line 42: Do you want to test how it handles tabs and quotes, if quotes are in the final output?

alxdarksage (approved):
Minor comment about testing input data with tabs

src/main/java/org/sagebionetworks/bridge/exporter/worker/TsvInfo.java
DwayneJengSage: Try-with-resources doesn't actually work here, because the CSVWriter is declared and created outside of the flushAndCloseWriter() method. I don't believe the code can be easily refactored to use try-with-resources, since the Writer can exist for a very long time and across multiple threads.

src/test/java/org/sagebionetworks/bridge/exporter/handler/PhoneAppVersionInfoTest.java
DwayneJengSage: No, because string sanitization occurs in another class, and I don't want to tightly couple the tests for this class with the implementation in a separate class.

PR #89, opened 2019-06-05
Schemaless Uploads

DwayneJengSage:
See https://sagebionetworks.jira.com/browse/BRIDGE-2466

Requires https://github.com/Sage-Bionetworks/BridgeServerLogic/pull/13

Note: We also remove an old hack for Share the Journey (breast cancer study). Since Share the Journey is closed, we don't need this hack, and also supporting this hack makes it more difficult to plumb the features we want.

### DIFF
--- a/src/main/java/org/sagebionetworks/bridge/exporter/handler/SchemalessExportHandler.java
+++ b/src/main/java/org/sagebionetworks/bridge/exporter/handler/SchemalessExportHandler.java
@@ -1,0 +1,51 @@
+package org.sagebionetworks.bridge.exporter.handler;
+
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+import org.sagebionetworks.bridge.exporter.metrics.Metrics;
+import org.sagebionetworks.bridge.exporter.synapse.SynapseHelper;
+import org.sagebionetworks.bridge.exporter.util.BridgeExporterUtil;
+import org.sagebionetworks.bridge.exporter.worker.ExportTask;
+import org.sagebionetworks.bridge.exporter.worker.MetaTableType;
+import org.sagebionetworks.bridge.exporter.worker.TsvInfo;
+import org.sagebionetworks.bridge.rest.model.UploadFieldDefinition;
+
+/** This worker exports schemaless records into the default table. */
+public class SchemalessExportHandler extends HealthDataExportHandler {
+    @Override
+    protected String getDdbTableName() {
+        return SynapseHelper.DDB_TABLE_SYNAPSE_META_TABLES;
+    }
+
+    @Override
+    protected String getDdbTableKeyName() {
+        return SynapseHelper.DDB_KEY_TABLE_NAME;
+    }
+
+    @Override
+    protected String getDdbTableKeyValue() {
+        return getStudyId() + "-default";
+    }
+
+    @Override
+    protected String getSynapseTableName() {
+        return BridgeExporterUtil.DEFAULT_TABLE_NAME;
+    }
+
+    @Override
+    protected TsvInfo getTsvInfoForTask(ExportTask task) {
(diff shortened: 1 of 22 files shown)


src/main/java/org/sagebionetworks/bridge/exporter/handler/SchemalessExportHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    protected String getDdbTableKeyValue() {
+        return getStudyId() + "-default";
+    }
+
+    @Override
+    protected String getSynapseTableName() {
+        return BridgeExporterUtil.DEFAULT_TABLE_NAME;
+    }
```
alxdarksage, line 35: They're going to add a schema later and then want all their original uploads to be in the new table. :-( No solution I can think of.

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -36,29 +34,11 @@
     public static final String CONFIG_KEY_TIME_ZONE_NAME = "time.zone.name";
     public static final String CONFIG_KEY_RECORD_ID_OVERRIDE_BUCKET = "record.id.override.bucket";
     public static final String CONFIG_KEY_SQS_QUEUE_URL = "exporter.request.sqs.queue.url";
-
+    public static final String DEFAULT_TABLE_NAME = "Default Health Data Record Table";
```
alxdarksage, line 37: "Schemaless Health Data Record Table" or "No Schema Health Data Record Table" ?

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
```diff
@@ -82,6 +62,11 @@
      * @return the schema key associated with that record
      */
     public static UploadSchemaKey getSchemaKeyForRecord(Item record) {
+        if (!record.hasAttribute("schemaId") || !record.hasAttribute("schemaRevision")) {
+            // This is a schemaless record.
+            return null;
+        }
```
alxdarksage, line 68: It would be nice during validation if we find schemaId without schemaRevision or surveyGuid without surveyCreatedOn, or vice versa, we logged a warning or something. It's good to persist the upload but having partial information seems to suggest that they are trying to use a schema but don't fully understand our system.

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/exporter/handler/SchemalessExportHandler.java
DwayneJengSage: 100% not supported, for all the same reasons why we implemented Schemaless in the first place.

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
DwayneJengSage: I feel like "Default" better conveys what the table is than "Schemaless". The idea is that, going forward, Schemaless is the norm rather than the exception. So in this case, it really is the default.

src/main/java/org/sagebionetworks/bridge/exporter/util/BridgeExporterUtil.java
DwayneJengSage: Going forward, schemaless will be the norm, not the exception. It wouldn't be "partial information", nor would it indicate that researchers "don't fully understand our system". It will be a fully supported first-class citizen and should be the default way our system is used.

With that in mind, I don't believe it's worth logging a warning over.

alxdarksage: Not if both are missing (this is normal, yes), but if one of the two values but not both are present.

PR #100, opened 2020-03-29
storage location doesn't need to be saved to project

DwayneJengSage:
I had previously misunderstood storage location settings to mean it had to be set on a project in order to be used. As it turns out, the storage location merely needs to exist, doesn't have to be set on a project. In fact, setting it on the project seems to cause a recurring issue where Synapse keeps defaulting to our read-only bucket.

This code changes removes the auto-creation of storage locations and gets the storage location ID from config.


src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseHelper.java
```diff
@@ -311,102 +312,6 @@ public static boolean isCompatibleColumn(ColumnModel oldColumn, ColumnModel newC
...
-            uploadDestinationListSetting.setLocations(ImmutableList.of(DEFAULT_STORAGE_LOCATION_ID,
-                    storageLocationId));
-            uploadDestinationListSetting.setProjectId(projectId);
-            uploadDestinationListSetting.setSettingsType(ProjectSettingsType.upload);
-            createProjectSettingWithRetry(uploadDestinationListSetting);
-        }
-
-        return storageLocationId;
-    }
-
```
alxdarksage, line 409: So basically, all Bridge projects use the same storage location in config, they don't vary by project?

alxdarksage (approved):

src/main/java/org/sagebionetworks/bridge/exporter/synapse/SynapseHelper.java
DwayneJengSage: Correct, all Bridge projects use the same storage location, because storage location is simply a mapping to an S3 bucket. (This is a bit of an over-simplification, but it's essentially true.)

